PR #46, opened 2024-12-27
Add config options to add and remove blocks as infusion stabilizers

### LINKED ISSUE #41: Add config options to add/remove infusion stabilizer blocks
* Adding infusion stabilizers, currently, requires modding to implement `IInfusionStabilizer`. This is not very accessible.
* There is no way to remove infusion stabilizers (e.g. skulls because they might be too-easily farmed in some packs).

* Add configurable `stabilizerAdditions` and `stabilizerExclusions` lists to add/remove infusion stabilizers.
* Entries in those lists should have some form akin to `modid:blockId[:metadata]`, where metadata is optional.
* If a block is on the additions list, the runic matrix should treat that block as a stabilizer.
* If a block that *does* implement `IInfusionStablizer` (e.g. crystal clusters) is on the exclusions list, the runic matrix should not treat it as a stabilizer.

Other add-ons/mod integrations, which are impractical.

**Additional context**
Suggested by @Midnight145

rndmorris:
Resolves #41


rndmorris:
Please squash merge

src/main/java/dev/rndmorris/salisarcana/config/modules/BugfixesModule.java
```diff
@@ -35,11 +35,11 @@ public BugfixesModule() {
                 ConfigPhase.EARLY,
                 "infernalFurnaceDupeFix",
                 "Fixes a smelting duplication glitch with the Infernal Furnace"),
-            integerInfusionMatrixMath = new ToggleSetting(
```
Midnight145, line 38: This option should still exist on the off chance the other option is disabled, but only enabled if both itself is enabled and the new option is disabled

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
```diff
@@ -0,0 +1,274 @@
...
+        final var block = resolveBlock(entry);
+        if (block == null) {
+            return false;
+        }
+        if (block == Blocks.air) {
+            LOG.error("Attempted to add Air to {}. For technical reasons, this is not allowed.", name);
+            return false;
+        }
+        final var mdSet = blockMap.computeIfAbsent(block, k -> new HashSet<>());
+        if (entry.metadata > -1) {
```
Midnight145, line 174: should also probably check for 32767 (OreDictionary.WILDCARD_VALUE or whatever the constant is) here and tryAddItem

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
```diff
@@ -0,0 +1,274 @@
...
+            final var result = new ParsedEntry();
+            result.modId = parts[0];
+            result.id = parts[1];
+
+            // beyond this point, everything is optional (even if it doesn't parse correctly)
+
+            if (parts.length > 2) {
+                final var metadata = IntegerHelper.tryParse(parts[2]);
+                if (metadata != null) {
+                    result.metadata = metadata;
```
Midnight145, line 250: might be worth just setting meta to 0 so you can get rid of some null checks elsewhere

src/main/java/dev/rndmorris/salisarcana/lib/InfusionMatrixLogic.java
```diff
@@ -0,0 +1,151 @@
...
+     * 
+     * @param world The world to check.
+     * @param x     The z coordinate of the matrix.
+     * @param y     The y coordinate of the matrix.
+     * @param z     The z coordinate of the matrix.
+     * @return The symmetry and list of pedestals that would influence the matrix.
+     */
+    public static MatrixSurroundingsResult checkMatrixSurroundings(World world, int x, int y, int z) {
+        final var matrix = new MatrixOrigin(world, x, y, z);
+        final var result = new MatrixSurroundingsResult();
```
Midnight145, line 34: Where's the definition for MatrixSurroundingResult? I'm on mobile so it's a pain for me to find anything.

Midnight145 (changes requested):

src/main/java/dev/rndmorris/salisarcana/config/modules/BugfixesModule.java
rndmorris: This isn't being removed, just internally renamed. It has the same key in the config file.

src/main/java/dev/rndmorris/salisarcana/lib/InfusionMatrixLogic.java
rndmorris: It's further down in the same class. Its definition is
```java
public static class MatrixSurroundingsResult {

    public final ArrayList<ChunkCoordinates> pedestals = new ArrayList<>();
    public int symmetry = 0;
}
```

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
rndmorris: `result.metadata` is `int` with a default of `-1`, not `Integer`. The result of `IntegerHelper.tryParse(...)` is only null in the event of a failed parse. No further null checking required.

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
rndmorris: I'm not following what that would accomplish in this case. The lookup logic currently doesn't have any support for oredict stuff, but I could look into adding it.

src/main/java/dev/rndmorris/salisarcana/config/modules/BugfixesModule.java
Midnight145: I mean having a config option that's _just_ the integer math instead of the full rework, unless that's still there and i'm just not seeing it which is also possible

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
Midnight145: The name is misleading, it's not actually oredictionary, it just allows for any metadata value and some people may use that assuming that leaving meta out defaults to 0, like with minetweaker scripts

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
Midnight145: Oh true, not really sure how I missed that. I really shouldn't do code reviews while tired

src/main/java/dev/rndmorris/salisarcana/lib/InfusionMatrixLogic.java
Midnight145: Ahh, I see it now, didn't realize it was a nested class and missed it on my phone

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
rndmorris: Ahhh, okay. Yeah, I'll do that then.

src/main/java/dev/rndmorris/salisarcana/config/modules/BugfixesModule.java
rndmorris: It's not. Having both the integer math and the full rework as options feels... I dunno, weird? Like, the integer math replaces just part of the method, but the rework replaces the whole thing *and* the rework

Midnight145: I think it's worth it because replacing the entire thing will kill any other modifications that any other mods do

rndmorris: Do you think requiring the add/remove stabilizers feature to depend on the rework is reasonable? Because otherwise there really aren't any good injection points in the first half of the method to make that work

Midnight145: Oh yeah absolutely. I was thinking to have the integer math as it's own option that is disabled if the add/remove feature is enabled as that'll replace it

src/main/java/dev/rndmorris/salisarcana/config/settings/BlockItemListSetting.java
rndmorris: Done. It now defaults to 0 if metadata isn't specified or couldn't be parsed. As a shorthand, `*` will now be interpreted as the oredict wildcard value.

src/main/java/dev/rndmorris/salisarcana/config/modules/BugfixesModule.java
rndmorris: Done. The rewrite is now tied into its own setting, and the integer math mixin has some logic on it to prevent it from loading alongside the rewrite mixin.

Midnight145 (approved):

PR #47, opened 2024-12-30
Prevent thaum items with invalid metadata from crashing clients

rndmorris:


rndmorris:
Actually, there are a few other items susceptible to the same `ArrayIndexOutOfBoundsException` crash if created with invalid metadata (`ItemNugget` and `ItemResource`). I'm tempted to just lump those in with this one under a single `itemMetadataCrash` setting.

Thoughts?

Midnight145:
Yeah that would work

What would honestly be easier/cleaner, if it's all exactly the same issue, is to just write an asm transformer for it, but that might not be worth the effort

Midnight145 (approved):
Approved, but I won't merge it yet in case you want to do the others under the same PR

rndmorris:
> Yeah that would work
> 
> What would honestly be easier/cleaner, if it's all exactly the same issue, is to just write an asm transformer for it, but that might not be worth the effort

So long as all of your target classes have the same shape you try to use in your mixin class (e.g. all have an `IIcon[] icon` field, all have a `getIconFromDamage(int)` method, etc), it's surprisingly flexible. It almost feels like it allows duck-typing.

Midnight145:
This likely needs Hodgepodge compat, from when I was looking earlier it fixes some of the same things, including the exact same duck typing solution for the iicon fix. I can double check here in a bit if you want

rndmorris:
Does it? Oops. The only part of Hodgepodge I've looked at was that link to its Mixins enum you sent me.

Midnight145 (approved):
Double checked hodgepodge, it changes something else that mixins in a bunch of classes, not this so we're good

PR #50, opened 2024-12-31
mixin compat via priority set, yoink hodgepodge mixin flow, yoink MixinTileWandPedestal

### LINKED ISSUE #4: Wand Recharge Pedestals should be able to use CentiVis
Normally Wand pedestals can only recharge vis receptacles from normal nodes. This is fine, but feels strange given that energized nodes exist.

Wand recharge pedestals should be able to recharge centivis, in addition to normal vis.

Another mod (I can't remember the name) adds similar behavior.

**Additional context**
Per S.O.P. this should be behind a config option, so that it can be disabled in the event the other mod is loaded (we could also check if said mod is loaded and disable automatically, if I can remember which one it is).

Midnight145:
closes #4

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
@@ -15,139 +15,138 @@ import java.util.function.Supplier;
 
 import cpw.mods.fml.relauncher.FMLLaunchHandler;
 import dev.rndmorris.salisarcana.config.ConfigModuleRoot;
-import dev.rndmorris.salisarcana.config.IEnabler;
-import dev.rndmorris.salisarcana.config.settings.Setting;
 
 public enum Mixins {
 
     // Morpheus
     BIOME_COLOR_EERIE_BASE_COLOR(new Builder().setPhase(Phase.LATE)
         .setSide(Side.BOTH)
-        .setConfig(ConfigModuleRoot.biomeColors.eerie.baseColor)
+        .applyIf(ConfigModuleRoot.biomeColors.eerie.baseColor::isEnabled)
         .addMixinClasses("world.biomes.eerie.MixinBaseColor")
         .addTargetedMod(TargetedMod.THAUMCRAFT)),
     BIOME_COLOR_EERIE_FOLIAGE_COLOR(new Builder().setPhase(Phase.LATE)
         .setSide(Side.BOTH)
-        .setConfig(ConfigModuleRoot.biomeColors.eerie.foliageColor)
+        .applyIf(ConfigModuleRoot.biomeColors.eerie.foliageColor::isEnabled)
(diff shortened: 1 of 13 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
```diff
@@ -1,127 +1,401 @@
...
+            if (this.phase != null) {
+                throw new RuntimeException("Trying to define Phase twice for " + this.name);
+            }
+            this.phase = phase;
+            return this;
+        }
+
+        public Builder setSide(Side side) {
+            if (this.side != null) {
+                throw new RuntimeException("Trying to define Side twice for " + this.name);
```
rndmorris, line 369: I can't say I'm thrilled at the overhead of having to pass in the string name of the enum the builder is for, but I'm not sure how else to catch duplicate side/phase assignments.

Maybe instead of throwing an error here, we could use lists inside the builder and throw an exception in the mixin constructor if it gets passed more than one side or phase?

rndmorris: It would remove the need for a "name" builder parameter, and group all of the runtime errors into one location

Midnight145: The name is just there as an identifier so you know _which_ mixin failed. We don't need the name at all, but we'd then have trouble finding which mixin threw the error.

rndmorris: My suggestion would still throw an error containing the name of the failed mixin, it would just throw the error from the enum constuctor (where the enum's name would natively accessible) instead of from the builder method.

Midnight145: How would you then determine if it was set twice? I don't _love_ the idea of a list, that's a lot of overhead. I guess we could just have an error flag that we check against.

Midnight145: Yeah, that just seems like a lot of overhead vs something like

```java
private boolean error = false;

public Builder setSide(Side side) {
    if (this.side != null) {
        error = true;
        return;
    }
    this.side = side;
}
```

```java
Mixins(Builder builder) {
    if (builder.side == null) {
        throw new RuntimeExpection("no side specified...
    } else if (builder.error) {
        throw new RuntimeException("multiple sides specified...
    }
}

rndmorris: I would be fine with that as well (though with `sideError` and `phaseError` fields, to communicate which one caused the failure)

Midnight145: Cool, sounds good

rndmorris (approved):

PR #53, opened 2025-01-02
Mod compatibility framework

Midnight145:
Adds a coremod to disable mods in the getMixins function for mods that don't have their mixins configurable, and adds a ModCompat class that can pull configuration options from mods to conditionally disable mixins based on what the other mod has enabled, instead of disabling them just if the mod is present at all.


src/main/java/dev/rndmorris/salisarcana/common/compat/ModCompat.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public static void init() {
+        // Both Hodgepodge and BugTorch register their configs in a coremod, so they are already loaded by the time
+        // we get here.
+
+        // For other mods, we may have to fetch config values manually.
+        if (Loader.isModLoaded("hodgepodge")) {
+            disableWandCV = TweaksConfig.addCVSupportToWandPedestal;
+            if (disableWandCV) {
+                LOG.info("Salis Arcana: Disabling Wand Pedestal CV support");
```
rndmorris, line 23: Suggestion: Put "to avoid conflicts" or something in these messages, since they're technically user-facing. I can feel at least one "Why is it disabling these configs I specifically enabled" issue in the future otherwise.

rndmorris (changes requested):

rndmorris (approved):

PR #56, opened 2025-01-09
Implement configurable infusion stabilizer strength options

### LINKED ISSUE #44: Add configurable infusion stabilizer strength, to let rarer stabilizers stabilize more
All stabilizers, regardless of rarity or difficulty to acquire, are exactly as strong as a little glob of magically-treated rotten flesh with string on top.

* Add an API to allow blocks that contribute a configurable amount of stability to an infusion altar.
* Compatibility with #41 can be accomplished with an optional fourth value `:strength` (being read as `modId:blockId:metadata:strength`, or `modId:blockId::strength` if metadata isn't specified).
* A `requireStrengthSymmetry` config option might be included to require a stabilizer and its pair to be equally as strong in order to effectively contribute to the altar's stability.
* For the lulz, allow negative strengths to create infusion *de*stabilizers

N/A

**Additional context**
Suggested by @Midnight145, among others

rndmorris:
Resolves #44

### DIFF
--- a/docs/enhancements.md
+++ b/docs/enhancements.md
@@ -78,6 +78,8 @@ Requires `useStabilizerRewrite=true`.
 
 The amount (in one-hundredths of a point) of symmetry each stabilizer block adds to an infusion altar. If a stabilizer doesn't have a symmetrical opposite, an equivalent amount of symmetry will be subtracted instead.
 
+Negative strengths will invert a stabilizer's behavior by stabilizing an infusion altar if placed asymmetrically, and destabilizing an altar if placed symmetrically.
+
 The default value of `10` is equivalent to vanilla Thaumcraft's stabilizer strength.
 
 ## Config option: `stabilizerAdditions`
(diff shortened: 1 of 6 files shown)


docs/enhancements.md
```diff
@@ -72,21 +72,39 @@ Rewrites the Runic Matrix's surroundings-check logic to be more flexible when ch
 This is a significant change, and will conflict with any other mod that also modifies the matrix's
 surroundings-checking logic.
 
+## Config option: `stabilizerStrength`
+
+Requires `useStabilizerRewrite=true`.
```
Midnight145, line 77: Do these settings also force useStabilizerRewrite if they are enabled but it isn't? If not, I think it should for ease of use, but I can see the argument for either way (since it's clear that it requires useStabilizerRewrite

rndmorris: They don't, no. They use `useStabilizerRewrite` as a dependency, and so inherit their own enabled status from it, but they don't update or override it themselves.

Since the rewrite is so invasive, I'm hesitant to let it be enabled by anything *but* a single, very explicit, opt-in setting.

Midnight145: Fair enough haha

Midnight145 (approved):

PR #59, opened 2025-01-10
Fix greatwood and silverwood slab burn times to match vanilla wooden slabs

rndmorris:


Midnight145 (approved):

rndmorris:
NEI shows the correct value, yes. I'm unsure about WAILA, but since we're using Forge's API I would hope WAILA would account for it.
image

Midnight145:
Ahh, alright, I assumed it was an event that was called when the furnace starts burning. I see how it works now.

Midnight145 (approved):

PR #82, opened 2025-01-14
Prevent a NullPointerException when building example wand replacement recipes

rndmorris:


src/main/java/dev/rndmorris/salisarcana/lib/WandHelper.java
```diff
@@ -109,32 +111,43 @@ public class WandHelper {
         return wandInstance;
     }
 
-    public static boolean isScepter(@Nullable ItemStack itemStack) {
```
rndmorris, line 112: The `isX` methods were unused, supplanted by the WandType enum, and the `primalAspectList` methods were unused duplicates of the ones in AspectHelper

Midnight145 (approved):

PR #86, opened 2025-01-14
Add config to remove gold coins from the oredict

### LINKED ISSUE #85: Toggle adding gold coins to the gold nugget oredict
Gold coins being in the gold nugget ore dictionary can cause balance problems in some situations, especially when oredict unifier blocks are present.

Add a config option to prevent gold coins from being oredicted.

MineTweaker may be able to remove it? I haven't checked

**Additional context**
This feels like it should go in enhancements, since being added to the oredict was an intentional decision. But the potential exploits this feature would prevent means maybe it should go in bugfixes.

rndmorris:
Resolves #85


Midnight145 (commented):
A mixin for this seems kinda heavy. If we already load after thaumcraft, why not just remove it from the oredictionary in postinit?

rndmorris:
If there's a public API through which to remove an OreDict entry, I haven't found it. Do you know of one?

Midnight145 (approved):

PR #96, opened 2025-01-18
configurable pure node biome changing to always change instead of just in blocks/tainted lands

Midnight145:


docs/enhancements.md
```diff
@@ -95,13 +95,21 @@ The weights used to pick the modifier of nodes when randomly generated.
...
+
 The weights used to pick the type of nodes when randomly generated.
 
 ## Mana Bean Growth Rate
 
 **Config Option:** `manaBeanGrowthChance`
+
 The random chance for mana beans to grow every block tick. Valid values are 0, 100 inclusive. Lower values are more likely to happen, with 0 being guaranteed.
 
+## Pure Node Biome Change Tweak
```
rndmorris, line 107: A more descriptive name ("Pure Nodes Always Create Magical Forest"?), for both the doc header and the actual config option name, would be preferred.

rndmorris (changes requested):

rndmorris (approved):

rndmorris (approved):

Issue #99, opened 2025-01-19
Eldritch Altar mob spawning logic is stupid


Nikolay-Sitnikov:
Whenever the eldritch altar spawns mobs (`TileEldritchAltar::spawnGuards()` and `TileEldritchAltar::spawnGuardian()`), the following code is used to choose a spawning location:
```java
int i1 = this.xCoord + MathHelper.getRandomIntegerInRange(this.worldObj.rand, 4, 10) * MathHelper.getRandomIntegerInRange(this.worldObj.rand, -1, 1);
int j1 = this.yCoord + MathHelper.getRandomIntegerInRange(this.worldObj.rand, 0, 3) * MathHelper.getRandomIntegerInRange(this.worldObj.rand, -1, 1);
int k1 = this.zCoord + MathHelper.getRandomIntegerInRange(this.worldObj.rand, 4, 10) * MathHelper.getRandomIntegerInRange(this.worldObj.rand, -1, 1);
```

The result of this integer math is that some spawning locations are more likely than others.
Specifically:
- `j1=0` 50% of the time, and is any other number 1/12th of the time.
- `i1` and `k1` are:  `0` at a 33% chance, `[-1,-3] ∪ [1, 3]` at a 0% chance, and `[-10, -4] ∪ [4, 10]` at a 1/21 chance each.

The result of this is:
- Hilly terrain makes the mobs spawn much slower
- There are areas in which mobs will never spawn (the center square, which is logical, and the cardinal strips going out from that square excluding the center lines, which isn't)
- Some areas have a much higher chance of spawning mobs than others
- 1/18 of all spawns try to spawn in the block space of the altar (and fail), while another 1/18 try to spawn directly above or below the altar (and also likely fail.)

rndmorris:
I'm open to making the mob spawns less insane, but please use the issue templates in the future.

PR #101, opened 2025-01-19
Provide more-sane eldritch altar mob-spawning behaviors

rndmorris:

New configurable feature

**What is the current behavior?**

See #99, which this closes.

Two new behaviors: even-spread (equal chance for each coordinate), and center-weighted (prefers near the altar). Neither try to spawn directly in the altar itself.

No

**Other information**:


src/main/java/dev/rndmorris/salisarcana/mixins/late/tiles/MixinTileEldritchAltar_EvenMobSpread.java
Midnight145, line 1: I don't *love* that they're two near-identical mixins, only differing by two unique functions, though I can't think of a "perfect" implementation otherwise without just checking the value every time.

Maybe when initializing the config, have a `Function<Random> randomVertical, randomHorizontal` and call xRef.set(x = MobSpawnDistributionSetting.randomHorizontal.apply(worldObj.rand));` (for example)?

You then still only have one function call so minimal runtime impact (especially in a case like this), and you avoid two nearly identical mixins which helps in case something ever needs changed in them.

Midnight145 (approved):

PR #103, opened 2025-01-20
misc. thaumometer enhancement fixes

### LINKED ISSUE #94: Scanning Gives No Research Points
**Describe the bug**
Scanning an item in your inventory gives no research points

**To Reproduce**
1. Check your number of research points
2. Scan a new item in your inventory
3. The item's aspects are learned, but no "ding" sound is played, no research point popup happens, and no research points get added

**Expected behavior**
Gain research points from scanned items

**Additional context**

Also happens when scanning an IInventory with new items inside

Midnight145:
closes #94

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/MixinScanManager.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/MixinScanManager.java
@@ -7,7 +7,6 @@ import net.minecraft.item.Item;
 import net.minecraft.tileentity.TileEntity;
 
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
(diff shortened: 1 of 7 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/MixinScanManager.java
```diff
@@ -1,18 +1,35 @@
...
+import thaumcraft.api.aspects.AspectList;
 import thaumcraft.api.research.ScanResult;
+import thaumcraft.common.lib.research.ResearchManager;
 import thaumcraft.common.lib.research.ScanManager;
 
 @Mixin(value = ScanManager.class, remap = false)
 public class MixinScanManager {
 
+    @Shadow
+    public static boolean isValidScanTarget(EntityPlayer player, ScanResult scan, String s) {
```
rndmorris, line 29: Since it's a public method, I think in this case it's fine to use `ScanManager.isValidScanTarget()` where this is needed. (I did test, and everything still works if you do).

Midnight145: Whoops, didn't realize that was static

rndmorris (approved):

src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/MixinScanManager.java
rndmorris: If it *wasn't* static, I would have told you to make your mixin and shadowed method abstract (which is what's recommended for mixin classes anyway).

Midnight145: Oh that's definitely what I meant to do either way, I think I pressed the wrong fix in Intellij to add a body instead of making it abstract

PR #104, opened 2025-01-20
notice if tcinventoryscan is independently installed

### LINKED ISSUE #95: Duplicate Inventory Scanning Tooltip
**Describe the bug**
Duplicate inventory scanning tooltip when tcinventoryscan is also present

**To Reproduce**
1. Have both salis arcana and tcinventoryscan installed
2. Look at the thaumometer

**Expected behavior**
The tooltip only shows up once. Maybe send a message to remove tcinventoryscan if it finds that it is also installed?

**Screenshots**

Image

**Additional context**
Maybe also add something that says about scanning inventory blocks with new items inside.

Midnight145:
closes #95 

image

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/notifications/StartupNotifications.java
+++ b/src/main/java/dev/rndmorris/salisarcana/notifications/StartupNotifications.java
@@ -2,23 +2,31 @@ package dev.rndmorris.salisarcana.notifications;
 
 import net.minecraft.util.ChatComponentTranslation;
 
+import cpw.mods.fml.common.Loader;
 import cpw.mods.fml.common.eventhandler.SubscribeEvent;
 import cpw.mods.fml.common.gameevent.PlayerEvent;
 import dev.rndmorris.salisarcana.config.ConfigModuleRoot;
 
 public class StartupNotifications {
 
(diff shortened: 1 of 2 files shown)


src/main/java/dev/rndmorris/salisarcana/notifications/StartupNotifications.java
```diff
@@ -2,23 +2,31 @@
...
 
 public class StartupNotifications {
 
-    private boolean noticeSent = false;
+    private boolean deprecationNoticeSent = false;
+
+    private boolean tcInventoryScanNoticeSent = !Loader.isModLoaded("tcinventoryscan");
 
     @SubscribeEvent
     public void onLogin(PlayerEvent.PlayerLoggedInEvent event) {
```
rndmorris, line 17: In this case, since we now have multiple notices, I think it would make sense to put each on in their own method, called from here. e.g.

```java
public void onLogin(...) {
    sendDeprecationNotice(event);
    sendTcInventoryScanNotice(event);
}
```

rndmorris (changes requested):

rndmorris (approved):

PR #107, opened 2025-01-20
fix staff foci effect offset

### LINKED ISSUE #97: Focus effects originate from the wrong place when using a staff
**Describe the bug**
Focus effects originate from the wrong place when using a staff

**To Reproduce**
1. Make a staff
2. Slap on a focus that spawns an effect (like excavation)
3. Use it

**Expected behavior**
The effect originates from the top of the staff instead of in midair

**Screenshots**
Wand:
Image
Staff:
Image

**List of affected foci**
Fire, shock, excavation, probably other modded foci with similar effects

Primal, pech's curse, and ice foci have a "swing" animation so they are less noticeable

Midnight145:
closes #97

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -82,7 +82,7 @@ usesMixins = true
 separateMixinSourceSet =
 
 # Adds some debug arguments like verbose output and class export.
-usesMixinDebug = false
+usesMixinDebug = true
 
 # Specify the location of your implementation of IMixinConfigPlugin. Leave it empty otherwise.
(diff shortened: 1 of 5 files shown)


gradle.properties
rndmorris, line 1: I'd prefer to leave this false by default

Midnight145: whoops, didn't actually mean to commit that, I've been intentionally not staging that

rndmorris (approved):

PR #110, opened 2025-01-24
Custom research creator and editor

Midnight145:


src/main/java/dev/rndmorris/salisarcana/lib/ResearchHelper.java
```diff
@@ -169,4 +175,52 @@ public static void completeResearchClient(EntityPlayer player, String research)
...
+            LOG.error("Research entry missing key.");
+            return false;
+        } else if (research.getCategory() == null || research.getCategory()
+            .isEmpty() || ResearchCategories.getResearchList(research.getCategory()) == null) {
+                LOG.error("Research entry {} category missing or invalid.", research.getKey());
+                return false;
+            }
+        ResearchItem original = ResearchCategories.getResearch(research.getKey());
+        if (original != null) {
+            research.updateResearchItem(original);
```
rndmorris, line 211: Error: updating the `category` field won't, on its own, change which page an existing research appears on. To fix this, remove the research from the original category, and put it in the new one. I.e.

```java
ResearchCategories.researchCategories.get(original.category).research.remove(original.key);
research.updateResearchItem(original);
ResearchCategories.researchCategories.get(original.category).research.put(original.key, original);
```

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
```diff
@@ -0,0 +1,471 @@
...
+            case NORMAL_CRAFTING:
+            case SMELTING: {
+
+                type = page.type == ResearchPage.PageType.CRUCIBLE_CRAFTING ? "crucible"
+                    : page.type == ResearchPage.PageType.INFUSION_CRAFTING ? "infusion"
+                        : page.type == ResearchPage.PageType.ARCANE_CRAFTING ? "arcane"
+                            : page.type == ResearchPage.PageType.SMELTING ? "smelting" : "crafting";
+                number = index;
+                item = new ItemEntry();
+                item.item = GameRegistry.findUniqueIdentifierFor(page.recipeOutput.getItem())
```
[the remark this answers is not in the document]
rndmorris: If needed, before a release, I can take a look at (de-)serializing recipes, myself.

src/main/java/dev/rndmorris/salisarcana/common/commands/CommandDumpResearch.java
```diff
@@ -0,0 +1,84 @@
...
+import dev.rndmorris.salisarcana.common.commands.arguments.ArgumentProcessor;
+import dev.rndmorris.salisarcana.common.commands.arguments.annotations.FlagArg;
+import dev.rndmorris.salisarcana.common.commands.arguments.annotations.NamedArg;
+import dev.rndmorris.salisarcana.common.commands.arguments.handlers.IArgumentHandler;
+import dev.rndmorris.salisarcana.common.commands.arguments.handlers.ResearchKeyHandler;
+import dev.rndmorris.salisarcana.common.commands.arguments.handlers.flag.FlagHandler;
+import dev.rndmorris.salisarcana.config.ConfigModuleRoot;
+import dev.rndmorris.salisarcana.lib.ResearchHelper;
+
+public class CommandDumpResearch extends ArcanaCommandBase<CommandDumpResearch.Arguments> {
```
rndmorris, line 24: Preference: Rename the command to `export-research` instead

src/main/resources/assets/salisarcana/lang/en_US.lang
```diff
@@ -102,6 +102,10 @@ salisarcana:command.upgrade-focus.noItem=Target player is not holding a wand foc
 salisarcana:command.upgrade-focus.tooManyUpgrades=%d upgrade(s) could not be applied due to the five-upgrade limit. 
 salisarcana:command.upgrade-focus.success=All focus upgrades applied successfully.
 
+salisarcana:commands.dump_research.success=§9Success:§r Dumped %s research to %s.
+salisarcana:commands.dump_research.usage=/salisarcana-dump-research --research <research-key> [--overwrite]
+salisarcana:commands.dump_research.failed=§cFailure:§r Failed to dump %s. Research likely has unsupported properties.
```
rndmorris, line 107: Preference: use the verb "export" instead of "dump"

src/main/resources/assets/salisarcana/lang/en_US.lang
```diff
@@ -102,6 +102,10 @@ salisarcana:command.upgrade-focus.noItem=Target player is not holding a wand foc
 salisarcana:command.upgrade-focus.tooManyUpgrades=%d upgrade(s) could not be applied due to the five-upgrade limit. 
 salisarcana:command.upgrade-focus.success=All focus upgrades applied successfully.
 
+salisarcana:commands.dump_research.success=§9Success:§r Dumped %s research to %s.
```
rndmorris, line 105: Preference: use the verb "export" instead of "dump"

src/main/java/dev/rndmorris/salisarcana/lib/ResearchHelper.java
```diff
@@ -169,4 +175,52 @@ public static void completeResearchClient(EntityPlayer player, String research)
                 player.worldObj.provider.dimensionId,
                 (byte) 0));
     }
+
+    public static String dumpResearchToJson(ResearchItem research) {
```
rndmorris, line 179: Preference: use the name `exportResearchToJson`

src/main/java/dev/rndmorris/salisarcana/lib/ResearchHelper.java
```diff
@@ -169,4 +175,52 @@ public static void completeResearchClient(EntityPlayer player, String research)
...
+            .create();
+        return gson.toJson(entry);
+    }
+
+    public static String dumpResearchToJson(String researchKey) {
+        ResearchItem research = ResearchCategories.getResearch(researchKey);
+        return dumpResearchToJson(research);
+    }
+
+    public static ResearchEntry loadResearchFromJson(String json) throws JsonSyntaxException {
```
rndmorris, line 194: Preference: use the name `importResearchFromJson`

rndmorris (changes requested):
In hindsight, I should have pushed for this to be broken up into two or three smaller, more discretely-reviewable PRs instead.

docs/research/research.md
```diff
@@ -0,0 +1,16 @@
+# Research Editor
+
+Salis Arcana includes a JSON-based research creator and editor, allowing you to create new researches and edit preexisting ones.
```
rndmorris, line 3: Request: "preexisting" should be "pre-existing"

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/network/MessageSendResearch.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public MessageSendResearch(ResearchEntry research) {
+        System.out.println("Sending research: " + research.getKey());
+        this.research = research;
+    }
+
+    @Override
+    public void fromBytes(ByteBuf buf) {
+        String json = new String(buf.array());
+        String parsed = json.substring(json.indexOf('{'), json.lastIndexOf('}') + 1); // remove any weird characters
```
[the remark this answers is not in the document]
Midnight145: Yeah, I was getting some weird invalid characters before the actual json itself

src/main/java/dev/rndmorris/salisarcana/config/modules/EnhancementsModule.java
```diff
@@ -204,13 +205,11 @@ public EnhancementsModule() {
...
                 "While in creative mode, ctrl + left click on a research in the Thaumonomicon to complete it."),
-            thaumometerScanContainersResearch = new CustomResearchSetting(
-                thaumometerScanContainers,
-                ConfigPhase.LATE,
-                "thaumometerScanContainers",
-                "Enable the thaumometer to scan the contents of inventories when right-clicking on them.",
-                new CustomResearchSetting.ResearchInfo("CHESTSCAN", "BASICS", 8, 3).setDifficulty(3)
-                    .setParents("DECONSTRUCTOR").setAspects("ordo:10", "perditio:10", "permutatio:10"))
+            thaumometerScanContainersResearch = new StringSetting(
+                this,
```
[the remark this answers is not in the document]
Midnight145: I may make a paired ToggleSetting for it so you can choose whether or not to disable the research

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
```diff
@@ -0,0 +1,471 @@
...
+    public ResearchEntry() {
+
+    }
+
+    public ResearchEntry(ResearchItem research) {
+        this.key = research.key;
+        this.category = research.category;
+        this.name = research.getName();
+        this.tooltip = research.getText();
+        this.aspects = new AspectEntry[research.tags.size()];
```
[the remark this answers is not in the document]
Midnight145: Thanks, my functional programming really isn't great, especially in Java

src/main/java/dev/rndmorris/salisarcana/common/commands/CommandDumpResearch.java
```diff
@@ -0,0 +1,84 @@
...
+                Files.write(Paths.get(file.toURI()), json.getBytes());
+            } catch (Exception e) {
+                LOG.error(e);
+                CommandErrors.generic();
+                return;
+            }
+
+            sender.addChatMessage(
+                new ChatComponentTranslation(
+                    "salisarcana:commands.dump_research.success",
```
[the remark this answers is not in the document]
Midnight145: Yeah no clue what's up with that

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemThaumometer.java
```diff
@@ -50,7 +50,7 @@ private boolean rescanInventory(EntityPlayer player, ScanResult item, String t,
```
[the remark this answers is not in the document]
Midnight145: The entire method isn't, only the first half to see if it should exit early

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
```diff
@@ -0,0 +1,471 @@
...
+            return new Aspect[0];
+        }
+        List<Aspect> aspectList = new ArrayList<>();
+        for (AspectEntry entry : aspectTriggers) {
+            aspectList.add(Aspect.getAspect(entry.getAspect()));
+        }
+        return aspectList.toArray(new Aspect[0]);
+    }
+
+    public void updateResearchItem(ResearchItem research) {
```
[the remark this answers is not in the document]
Midnight145: My solution *should* work now, though if some mod does something super weird like text.key.somethingelse I should be able to change it to substr(0, indexof(key)) + key + substr( indexof(key), str.length)) or whatever

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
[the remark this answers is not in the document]
Midnight145: Yeah, I'm not sure how to best handle multi-page recipes and things with "fuzzy" item outputs

src/main/java/dev/rndmorris/salisarcana/common/commands/CommandDumpResearch.java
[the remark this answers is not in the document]
Midnight145: actually I think that second dot is just part of the j because there's no kerning

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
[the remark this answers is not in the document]
Midnight145: Has now been implemented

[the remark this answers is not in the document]
Midnight145: actually might not work for TT, though I might just mixin to tinkerer and @Overwrite the method to return `ttresearch.name`, `ttresearch.text`, etc.

It should work in most cases though, TT just pre-translates all of its keys

src/main/java/dev/rndmorris/salisarcana/common/commands/CommandExportResearch.java
```diff
@@ -41,42 +44,43 @@ protected int minimumRequiredArgs() {
     }
 
     @Override
-    protected void process(ICommandSender sender, CommandDumpResearch.Arguments arguments, String[] args) {
-        for (String researchKey : arguments.researchKeys) {
-            String json;
+    protected void process(ICommandSender sender, CommandExportResearch.Arguments arguments, String[] args) {
+        Path path = Paths.get("config/salisarcana/research/export");
```
rndmorris, line 48: My suggestion about using `Paths.get()` was more about using it to ensure the correct file separator character (`\` on Windows, `/` elsewhere) gets used. As-is, I believe this would cause problems if run on Windows.

src/main/java/dev/rndmorris/salisarcana/common/commands/CommandExportResearch.java
```diff
@@ -41,42 +44,43 @@ protected int minimumRequiredArgs() {
...
+                LOG.error(e);
+                CommandErrors.generic();
+                return;
             }
-
-            File file = new File("config/salisarcana/research/" + researchKey + ".json");
+        }
+        for (ResearchItem research : arguments.researches) {
+            String researchKey = research.key;
+            File file = Paths.get("config/salisarcana/research/export/" + researchKey + ".json")
```
rndmorris, line 60: Likewise here

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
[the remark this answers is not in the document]
rndmorris: If you look at `ResearchItem#getName()`, TT is doing the exact same thing TC is doing (just with a different key structure).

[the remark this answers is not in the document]
rndmorris: I would prefer we not mixin more mods than we need.

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
```diff
@@ -174,29 +179,40 @@ public ItemStack[] getItemTriggers() {
...
-        AssetHelper.addLangEntry("tc.research_name." + this.getKey(), this.getName());
-        AssetHelper.addLangEntry("tc.research_text." + this.getKey(), this.getTooltip());
+        if (!research.category.equals(this.getCategory())) {
+            r.set("category", this.getCategory());
+            ResearchCategories.researchCategories.get(research.category).research.remove(research.key);
+            ResearchCategories.researchCategories.get(research.category).research.put(research.key, research);
+        }
+        String name = research.getName();
+        String text = research.getText();
+        String parsed = name.substring(0, name.indexOf(key)) + name.substring(name.indexOf(key));
```
rndmorris, line 212: I don't think this works the way you think it does (see previous comment about the definition of `ResearchItem#getName()` and `ResearchItem#getText()`).

src/main/java/dev/rndmorris/salisarcana/lib/AssetHelper.java
```diff
@@ -52,12 +54,12 @@ public static void copyResearchFiles() {
                 return;
             }
         }
-        copyFiles(RESOURCE_PATH, CONFIG_PATH);
+        extractFilesFromResources(RESOURCE_PATH, CONFIG_PATH);
```
rndmorris, line 57: Again, we might as well just move these constants into the only method using them.

src/main/java/dev/rndmorris/salisarcana/network/MessageSendResearch.java
```diff
@@ -14,6 +14,16 @@
 public class MessageSendResearch implements IMessage, IMessageHandler<MessageSendResearch, IMessage> {
 
     private ResearchEntry research;
+    private static Gson gson = new Gson();
```
rndmorris, line 17: This should either be `null` by default, or `final` without `getGson()`.

rndmorris (changes requested):

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+      "type": "integer",
+      "description": "The complexity of this research. Used in creating research notes."
+    },
+    "isSpecial": {
+      "type": "boolean",
+      "description": "Whether this research is \"special\". Slightly changes the icon of the research. Mutually exclusive with isRound."
+    },
+    "isSecondary": {
+      "type": "boolean",
+      "description": "Whether this research is secondary"
```
rndmorris, line 84: If we know what "secondary" actually means, it would be helpful to include it here.

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+    },
+    "isRound": {
+      "type": "boolean",
+      "description": "Whether this research is round. Changes the icon of the research. Mutually exclusive with isSpecial."
+    },
+    "isStub": {
+      "type": "boolean",
+      "description": "Whether this research is a stub"
```
rndmorris, line 92: Likewise, if we know what this means we should include it.

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+    },
+    "isVirtual": {
+      "type": "boolean",
+      "description": "Whether this research is virtual. This will not be displayed in the Thaumonomicon, but can be used as a parent for other researches"
+    },
+    "isConcealed": {
+      "type": "boolean",
+      "description": "Whether this research is concealed until parents are completed."
```
rndmorris, line 100: And likewise what the differences between "hidden" and "concealed" are.

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+      "type": "boolean",
+      "description": "Whether this research is hidden until parents are completed."
+    },
+    "isLost": {
+      "type": "boolean",
+      "description": "Whether this research is \"lost\". Seems to have something to do with Eldritch research"
+    },
+    "isAutoUnlock": {
+      "type": "boolean",
+      "description": "Whether this research is auto-unlocked on completion of parents"
```
rndmorris, line 112: Unless my memory is faulty, this is untrue. Auto-unlocked is *always* unlocked, not just when its parents are completed.

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+    },
+    "itemTriggers": {
+      "type": "array",
+      "description": "The item triggers for this research",
```
rndmorris, line 116: Here, and `entityTriggers` and `aspectTriggers`, it's worth noting what this does (reveal hidden research).

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+      "description": "The aspect triggers for this research",
+      "items": {
+        "type": "string"
+      },
+      "uniqueItems": true
+    },
+    "pages": {
+      "type": "array",
+      "description": "The pages for this research",
+      "items": {
```
rndmorris, line 141: Instead of a universal catch-all object definition, we should have definitions for each page type that only contain the fields we accept for that type.

docs/research/schematic.json
```diff
@@ -0,0 +1,211 @@
...
+        "type": "object",
+        "properties": {
+          "pageType": {
+            "type": "string",
+            "description": "The type of the page",
+            "enum": ["text", "arcane", "recipe", "image", "smelting", "crafting", "infusion"]
+          },
+          "number": {
+            "type": "integer",
+            "description": "The page number. Must start at 0 and increase by 1 for each page",
```
rndmorris, line 151: If it *must* start at 0, and always increase by 1, why aren't we using the array index on our end?

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/lib/ResearchHelper.java
```diff
@@ -169,4 +191,51 @@ public static void completeResearchClient(EntityPlayer player, String research)
                 player.worldObj.provider.dimensionId,
                 (byte) 0));
     }
+
+    public static void exportResearchToJson(@Nonnull ResearchItem research, File dumpTo) throws IOException {
```
rndmorris, line 195: This is on me, but this should be `ResearchEntry` not `ResearchItem`. As-is it exports the research internals, which is not what we want.

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/common/commands/CommandExportResearch.java
```diff
@@ -0,0 +1,88 @@
...
+    public CommandExportResearch() {
+        super(ConfigModuleRoot.commands.exportResearch);
+    }
+
+    @Override
+    protected @Nonnull ArgumentProcessor<CommandExportResearch.Arguments> initializeProcessor() {
+        return new ArgumentProcessor<>(
+            CommandExportResearch.Arguments.class,
+            CommandExportResearch.Arguments::new,
+            new IArgumentHandler[] { ResearchKeyHandler.INSTANCE, FlagHandler.INSTANCE });
```
rndmorris, line 38: `ResearchKeyHandler` needs to be `ResearchHandler`. As-is, this crashes at startup.

rndmorris:
Reminder for @rndmorrs: the chest scan research, though enabled, is not getting registered as of the latest commit.

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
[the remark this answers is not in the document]
rndmorris: Though replacing the original ResearchItem instance with the default one is probably a bad idea as well. Automagy does some stuff inside its own derivative class, and replacing the object would risk breaking that.

docs/research/schematic.json
Midnight145: I'm not entirely sure how to do that tbh, this is my first time with json schemas and am completely winging it

docs/research/schematic.json
Midnight145: I know I definitely had a reason, though I'm not sure if I remember it.

Midnight145:
> Reminder for @rndmorrs: the chest scan research, though enabled, is not getting registered as of the latest commit.

fixed

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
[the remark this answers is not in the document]
Midnight145: fixed by iterating over lang dict values

src/main/java/dev/rndmorris/salisarcana/lib/AssetHelper.java
```diff
@@ -39,6 +40,20 @@ public class AssetHelper {
             .get(languageList, Map.class);
     }
 
+    public static String lookupLangEntryByValue(String value) {
+        ArrayList<String> keys = new ArrayList<>();
+        LANGUAGE.entrySet()
```
rndmorris, line 45: Using a list here is somewhat wasteful. Try this instead. It still finds the first entry that meets the criteria, but stops as soon as we find it. Unfortunately we do have to use an array because `findFirst()` isn't available in Java 8 (as far as I know), but it'll be at most of length one.

```java
final var keys = LANGUAGE.entrySet()
    .stream()
    .filter(
        e -> e.getValue()
            .equals(value))
    .map(Map.Entry::getKey)
    // In almost all cases, there should only be one key for a given value
    .limit(1)
    .toArray(String[]::new);
return ArrayHelper.tryGet(keys, 0).data();
```

docs/research/schematic.json
rndmorris: It's been a while since I've had to do it. I can take a look at doing it, when I get a chance.

docs/research/schematic.json
```diff
@@ -0,0 +1,210 @@
...
+    },
+    "displayColumn": {
+      "type": "integer",
+      "description": "The column to display this research in."
+    },
+    "displayRow": {
+      "type": "integer",
+      "description": "The row to display this research in."
+    },
+    "icon_item": {
```
rndmorris, line 65: Likewise, there's a way make this mutually exclusive with `icon_resource` within the schema itself. I can look into getting that done.

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
```diff
@@ -278,6 +285,7 @@ public int getAmount() {
 
 class ResearchPageEntry {
 
+    @SerializedName("pageType")
```
rndmorris, line 288: As-is this violates both the schema and `ContainerScan.json`

Midnight145: Schema is correct, I just only updated the real config file instead of the bundled one, whoops

docs/research/schematic.json
Midnight145: That'd be great, if you don't mind.

docs/research/schematic.json
rndmorris: From outside discussions: This was intended to be part of allowing updating only pages, but that functionality was left out of the PR. We'll leave this here for now, and I'll update it appropriately in my follow-up PR.

docs/research/schematic.json
rndmorris: This will be addressed in a follow-up PR.

docs/research/pages.md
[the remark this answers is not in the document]
rndmorris: This will be addressed in a follow-up PR.

docs/research/schematic.json
rndmorris: This will be addressed in a follow-up PR.

src/main/java/dev/rndmorris/salisarcana/config/settings/ResearchEntry.java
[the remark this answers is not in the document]
rndmorris: This will be addressed in a follow-up PR.

rndmorris (approved):
Remaining concerns will be addressed in a follow-up PR.

rndmorris:
Before merging: this needs tested again to ensure it both doesn't crash and behaves as expected.

Midnight145:
> Before merging: this needs tested again to ensure it both doesn't crash and behaves as expected.

Should be good now

rndmorris:
I'm using a version of ContainerScan.json from before you updating page to pageType, and it crashes the game completely when it hits an NPE. Last thing I think we need for this one is a general catch around the parsing/updating process, just so we don't burn out entirely if something goes wrong.

rndmorris (approved):

rndmorris (approved):

PR #111, opened 2025-01-24
unset remap=false for vanilla methods so gradlew runObfClient doesn't crash

Midnight145:


src/main/resources/assets/salisarcana/research/ContainerScan.json
rndmorris, line 1: Looks like you accidentally checked in an extra file

rndmorris (approved):

PR #112, opened 2025-01-24
Replace injections and redirects with wraps where possible

rndmorris:


rndmorris:
Note-to-self: Beacon block fix needs an explicit enable/disable option

PR #130, opened 2025-02-16
Fix remap mistakes

rndmorris:


gradle.properties
```diff
@@ -82,7 +82,7 @@ usesMixins = true
 separateMixinSourceSet =
 
 # Adds some debug arguments like verbose output and class export.
-usesMixinDebug = false
+usesMixinDebug = true
```
rndmorris, line 85: Setting this to true by default so that we can catch these mistakes sooner

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemWandCasting_DisableSpendingCheck.java
```diff
@@ -12,7 +12,7 @@ public class MixinItemWandCasting_DisableSpendingCheck {
 
     @ModifyExpressionValue(
         method = "consumeAllVis",
-        at = @At(value = "FIELD", target = "Lnet/minecraft/world/World;isRemote:Z"))
+        at = @At(value = "FIELD", target = "Lnet/minecraft/world/World;isRemote:Z", remap = true))
```
Midnight145, line 15: since this is the only method in the class, we should just remove the remap=false from the Mixin annotation

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemWandCasting_NamedStaffters.java
```diff
@@ -21,7 +21,11 @@ public abstract class MixinItemWandCasting_NamedStaffters {
     @Shadow(remap = false)
```
Midnight145, line 21: this and the other variable don't need remap=false anymore since we added it to mixin annotation

Midnight145 (changes requested):

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemWandCasting_DisableSpendingCheck.java
rndmorris: We can remove it from there, but it will need added to the `@ModifyExpressionValue` properties, since `consumeAllVis` is not a vanilla method.

Midnight145: Oop, yeah, that's my bad, sorry about that

Midnight145 (approved):

PR #135, opened 2025-02-17
add type adapters for research page entries

Midnight145:


src/main/java/dev/rndmorris/salisarcana/lib/customresearch/pages/CraftingResearchPageEntry.java
```diff
@@ -0,0 +1,50 @@
...
+        this.item = new ItemEntry(page.recipeOutput);
+    }
+
+    @Override
+    public ResearchPage getPage() {
+        ItemStack stack = item.getItemStack();
+        if (stack == null) {
+            return null;
+        }
+        for (Object obj : CraftingManager.getInstance()
```
rndmorris, line 30: I'm not sure why you're doing this when the recipe(s) actually on the page are stored in `ResearchPage.recipe`. It'd take some runtime type checking, but they're all there.

src/main/java/dev/rndmorris/salisarcana/lib/customresearch/pages/CraftingResearchPageEntry.java
```diff
@@ -0,0 +1,50 @@
...
+public class CraftingResearchPageEntry extends RecipeResearchPageEntry {
+
+    public CraftingResearchPageEntry() {
+        super();
+    }
+
+    public CraftingResearchPageEntry(ResearchPage page, Integer index) {
+        this.type = "crafting";
+        this.number = index;
+        this.item = new ItemEntry(page.recipeOutput);
```
rndmorris, line 21: `page.recipeOutput` is only useful for single-recipe pages. Pages can have more than one recipe, which they cycle through.

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/lib/customresearch/pages/CraftingResearchPageEntry.java
Midnight145: Yeah, I saw that while continuing work on recipe deserialization. That'll be addressed in a subsequent PR that I'm in the middle of

src/main/java/dev/rndmorris/salisarcana/lib/customresearch/pages/CraftingResearchPageEntry.java
Midnight145: This entire block gets rewritten once actual serialization/deserialization is done, so will also be fixed in the next PR

PR #145, opened 2025-02-19
Allow Wand Replacement Recipes to Complete Using the Wand Being Modified

Nikolay-Sitnikov:
It's only slightly hacky...

(Note: this technically does have one flaw. If you set up a replacement recipe in the GUI, and then a Dispenser equips you with Vis Discount gear, that discount won't apply until you move an item in the GUI. If you try to craft immediately after, you will not gain the benefits of the gear you have equipped.)

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/gui/MixinGuiArcaneWorkbench_SingleWandReplacement.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/gui/MixinGuiArcaneWorkbench_SingleWandReplacement.java
@@ -59,8 +59,7 @@ public class MixinGuiArcaneWorkbench_SingleWandReplacement {
         method = "drawGuiContainerBackgroundLayer",
         at = @At(
             value = "INVOKE",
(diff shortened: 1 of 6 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/gui/MixinGuiArcaneWorkbench_SingleWandReplacement.java
rndmorris, line 1: This works in singleplayer, but causes a client-side crash when you try to open the gui in multiplayer. Server-side appears unaffected. Relevant log snippet attached.

[fml-client-latest.log](https://github.com/user-attachments/files/18872400/fml-client-latest.log)

To reproduce:
1. Use the Run Obfuscated Server and Run Obfuscated Client run configurations concurrently. Respond to server prompts if needed.
2. In the client, connect to the server. Put down an arcane workbench, then right-click it.
3. See the crash.

rndmorris (changes requested):

Issue #30, opened 2024-12-20
Commands that print research names do not work for Thaumic Tinkerer


rndmorris:
**Describe the bug**
Thaumic Tinkerer overrides `ResearchItem::getName()` to use a different langfile key than vanilla Thaumcraft.

**To Reproduce**
1. Install Thaumic Tinkerer
2. Launch the game, load a save
3. Run `/list-research`
4. All of Thaumic Tinkerer's research will only print (invalid) langkeys

**Expected behavior**
The researches' actual names will be displayed

**Screenshots**
![image](https://github.com/user-attachments/assets/50bddfe9-f828-4291-9621-f2af0a701d59)

**Additional context**
May also affect Automagy

rndmorris:
Resolved with #153 (additional addons will be handled on a per-request case-by-case basis)

PR #147, opened 2025-02-23
Solution to #30 by dynamically adding extra translations

Nikolay-Sitnikov:
Solves #30 by adding new translation keys for any researches missing keys.

Several questions remain:
- Should I attempt to run a reverse lookup on the research keys not matching the pattern, or just let them go?
- Should I have put the dynamically generated keys in `tc.research_name.KEY` instead to force every mod to follow the pattern?
- Should I be keeping a cache of the `ResearchItem` translation keys (which can get large, since the client will cache every single research key because of the `ResearchTranslationGenerator`) or should I not keep a cache?


Nikolay-Sitnikov:
By answering "yes" to question 2, this solution becomes really overcomplicated and impractical. I'll write a much better version of this soon.

PR #156, opened 2025-02-28
Fix for #30 and Translation Manager

Nikolay-Sitnikov:
My take on the solution to the `getName` problem.
Originally, the plan was to override `getName` so it would preferentially use the standard lang key (`tc.research_name.KEY`) over whatever logic the addons had.

I removed my attempts at solving the problems. What's left is a tool for setting language entries across languages and across resource pack reloads.

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/lib/TranslationManager.java
+++ b/src/main/java/dev/rndmorris/salisarcana/lib/TranslationManager.java
@@ -69,13 +69,12 @@ public final class TranslationManager {
         }
     }
 
-
     /***
      * Sets a language key and value pair for en_US.
      * Automatically updates the language table when appropriate.
      *
-     * @param key      The translation key to create or update.
-     * @param value    The translated text to set the key to.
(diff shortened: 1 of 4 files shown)


rndmorris:
Please merge dev into your branch

Nikolay-Sitnikov:
Merged dev into branch and fixed all errors resulting from this change.

rndmorris:
Is this actually being used for anything now?

Nikolay-Sitnikov:
This is code which allows us to add new lang file entries dynamically while handling multiple languages.

It will almost certainly be useful if we plan to continue developing the custom research system.

---

I also planned to use this code to solve the research `getName` issue by mixing into all broken mods and forcing them to use the default translation keys, but I removed that part of the code from this PR after I heard you already solved that problem via a different route.

rndmorris:
Since it's not in use by anything right now, and would be essentially dead code, I'm going to close this without merging it.

That said, if we come across a situation where this would be a good solution, let's discuss getting this merged in alongside the code that would use it.

PR #157, opened 2025-03-02
Add recipes for removing upgrades from wand foci

### LINKED ISSUE #151: [Enhancement]: Focus Upgrade Removal
### Salis Arcana Version

v1.1.1

### Context

Upgrading foci without doing it in a test world first is always a gamble. Needing to make a new focus sucks (especially when they have complex recipes) when upgrades applied earlier are no longer useful (like Frugal).

### Requested Feature

Some way to remove the most recent upgrade from a focus in survival. Some xp being refunded based on the rank could be a nice bonus.

### Alternative Solutions

Maybe some way to entirely wipe a focus's upgrades and start from a fresh one without needing to make a whole new one could work, but it would be a bit less flexible than just removing the most recent upgrade.

### Additional Context

_No response_

rndmorris:
To resolve #151. I'm disinclined to implement XP refunds in addition to effectively refunding the materials used in the focus.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -39,7 +39,7 @@ dependencies {
     api("com.github.GTNewHorizons:Baubles:1.0.4:dev")
     api("thaumcraft:Thaumcraft:1.7.10-4.2.3.5:dev")
     compileOnly(rfg.deobf("curse.maven:tc4tweaks-431297:6019390"))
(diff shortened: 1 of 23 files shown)


rndmorris:
Please hold off on merging for a bit, there's a small bit more of cleanup that needs done

PR #163, opened 2025-03-08
custom thaumometer scan duration, closes #155

### LINKED ISSUE #155: [Enhancement]: Add config file option for players to configure Thaumometer Scan time
### Salis Arcana Version

None

### Context

As this is taking TCInventoryScan under its wing in GTNH 2.8.x, and TCIS is an archived mod, I have been instructed to write this feature request over here and  am writing this feature request over here. 

When scanning lots of items I am always frustrated b y how long I have to hover ocver each one, which led me to want to modify the SCAN_TICKS. As the developer(s) of this mod however did not add a config file (option) for this, I had to clone the github repo, find the value, change it, do ./gradlew copmile or smth and only then was able to set the SCAN_TICKS it took to a lower value. 
I am personally sure that other people have also been frustrated by this, and as a) not everyone is able to compile mods by themselves and b) this mod is gonna be a whole lot bigger, meanin that finding the relevant piece of code to chaneg will be a lot more time consuming, making that value configurable from a config file would only help users of this mod and make lowering the scan time much more accessible., it's not like that is a bigger change than maybe 12 lines of code in order to read a value from a config file. 

### Requested Feature

I want the SCAN_TICKS that it takes to scan items when hovering over them with the Thaumometer to be configurable from a config file in the instances config folder. 

### Alternative Solutions

Cloning the repo, editing the value, compiling the mod yourself and then sawpping it with the pack's version does work but, as I said, not everyone is able or will be able to do that. 

### Additional Context

_No response_

Midnight145:


src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
```diff
@@ -350,7 +352,15 @@ public enum Mixins {
         .setSide(Side.BOTH)
         .setApplyIf(ConfigModuleRoot.enhancements.primalCrusherOredict::isEnabled)
         .addMixinClasses("items.PrimalCrusher_StoneOredictCompat")
-        .addTargetedMod(TargetedMod.THAUMCRAFT));
+        .addTargetedMod(TargetedMod.THAUMCRAFT)),
+    THAUMOMETER_CUSTOM_DURATION(new Builder().setApplyIf(() -> ConfigModuleRoot.enhancements.thaumometerDuration.getValue() != ConfigModuleRoot.enhancements.thaumometerDuration.getDefaultValue())
```
rndmorris, line 356: I could have sworn I did this in the item decay PR (but I guess didn't commit it): Instead of this, we can override `IntSetting.isEnabled()` to `return value != defaultValue && super.isEnabled();`.

src/main/java/dev/rndmorris/salisarcana/client/ThaumicInventoryScanner.java
```diff
@@ -44,7 +45,11 @@
 
 public class ThaumicInventoryScanner {
 
-    private static final int SCAN_TICKS = 50;
+    // SCAN_TICKS is seemingly double the "vanilla" scan duration, I'm not sure if onClientTick is called twice as often
+    // or what, but it seems to work if I do this
+    // +5 is added to the duration to account for the fact that in the config, we set the duration to be the total usage
+    // duration - 5 since thaumcraft completes the scan when the remaining duration is <= 5
+    private static final int SCAN_TICKS = (ConfigModuleRoot.enhancements.thaumometerDuration.getValue() + 5) * 2;
```
rndmorris, line 52: Suggestion: This should probably be a (memoized) getter instead. Since it's a static field, it'll be initialized as soon as `ThaumicInventoryScanner` is loaded. I can't think of a situation where that would happen before we're done reading config values, but I'd prefer to be safe.

rndmorris (changes requested):

rndmorris (approved):

PR #164, opened 2025-03-09
fix: several config entries loaded in wrong phase

Midnight145:


Midnight145:
closing in favor of removing ConfigPhase entirely

PR #160, opened 2025-03-04
Add Focus Disenchanting

Nikolay-Sitnikov:
Closes #151.

Adds a new fake focus upgrade - "Disenchant Focus" and captures it in many places in order to make the GUI and the server work properly.

This PR has been tested in obfuscated singleplayer and obfuscated multiplayer.

Known issues:
- On multiplayer, using `/salisarcana-forget-research` doesn't update the client-side completed research list, thus letting the user keep seeing the "Disenchant Focus" option even though they are not able to use it.
- An error keeps popping up apparently related to the fact that one of the GTNH mod files is broken. I don't think that's my fault, I might try cleaning and rebuilding the cache after this. (See error below.)
- Using `/salisarcana-upgrade-focus` tab completion spams `NullPointerException`s. Again, don't think that's my fault, and everything still works correctly nonetheless.

```
[23:05:11] [Client thread/ERROR] [FML]: Unable to read a class file correctly
java.lang.IllegalArgumentException
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:170) ~[asm-debug-all-5.0.3.jar:5.0.3]
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:153) ~[asm-debug-all-5.0.3.jar:5.0.3]
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:424) ~[asm-debug-all-5.0.3.jar:5.0.3]
	at cpw.mods.fml.common.discovery.asm.ASMModParser.<init>(ASMModParser.java:52) [ASMModParser.class:?]
	at cpw.mods.fml.common.discovery.JarDiscoverer.discover(JarDiscoverer.java:72) [JarDiscoverer.class:?]
	at cpw.mods.fml.common.discovery.ContainerType.findMods(ContainerType.java:42) [ContainerType.class:?]
	at cpw.mods.fml.common.discovery.ModCandidate.explore(ModCandidate.java:71) [ModCandidate.class:?]
	at cpw.mods.fml.common.discovery.ModDiscoverer.identifyMods(ModDiscoverer.java:131) [ModDiscoverer.class:?]
	at cpw.mods.fml.common.Loader.identifyMods(Loader.java:364) [Loader.class:?]
	at cpw.mods.fml.common.Loader.loadMods(Loader.java:489) [Loader.class:?]
	at cpw.mods.fml.client.FMLClientHandler.beginMinecraftLoading(FMLClientHandler.java:208) [FMLClientHandler.class:?]
[...]

PR #167, opened 2025-03-09
Make /forget-research actually remove warp & synchonize changes in multiplayer

Nikolay-Sitnikov:
Fixed bugs:
- `/salisarcana-forget-research` didn't remove warp when command completed.
- `/salisarcana-forget-research` wouldn't synchronize research to the client, which caused bugs like flickering and ghost items in the Arcane Workbench, or the Disenchant Focus icon being visible in #160.


Midnight145 (approved):

rndmorris:
Not removing warp wasn't a bug, it was intentional design.

Nikolay-Sitnikov:
> Not removing warp wasn't a bug, it was intentional design.

Then the success message shouldn't mention the total amount of warp (which is implied to be removed along with the research.) If you want, I can add a setting to disable removing warp.

PR #181, opened 2025-03-17
Use Forge's fishing lists for fishing golem catch lists

rndmorris:
Bug fix (sorta)

Fishing golem AI draw catches from their own loot tables.

The fishing golem AI's loot tables are replaced with Forge's fishing loot tables.

Not that I'm aware of

**Other information**:


src/main/java/dev/rndmorris/salisarcana/CommonProxy.java
```diff
@@ -91,4 +104,30 @@ private void maybeRegister(FMLServerStartingEvent event, CommandSettings setting
     public boolean isSingleplayerClient() {
         return false;
     }
+
+    private void fixGolemFishingLists() {
+        try {
+            final var fishingHooks = new R(FishingHooks.class);
+            final var lootCrap = AIFish.class.getDeclaredField("LOOTCRAP");
```
Midnight145, line 111: R has a method that would let you do something like `R.of(AIFish.class).set("LOOTCRAP", fishingHooks.get("junk", ArrayList.class));` to avoid having to handle field protection manually

rndmorris: That was the first thing I tried, but R doesn't handle `final` fields, which TC's lists are.

Midnight145: Weird, docstring says it can. I may modify R so it unsets the final field later, but for now this is good.

Midnight145 (approved):

PR #194, opened 2025-03-20
R::set now works with final parent fields, don't lookup toSet twice

Midnight145:
R::set now works with parent fields, don't lookup toSet twice

r doesn't set final modifier on parent fields, and looks up toSet twice which is unnecessary

R::set now works with final parent fields, and doesn't lookup toSet twice

no

**Other information**:

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/lib/R.java
+++ b/src/main/java/dev/rndmorris/salisarcana/lib/R.java
@@ -95,9 +95,9 @@ public class R {
     public R set(String name, Object value) {
         try {
             Field toSet = findField(name, clazz);
-            Field modifiersField = findField("modifiers", Field.class);
+            Field modifiersField = findField("modifiers", toSet.getClass());
             modifiersField.setInt(toSet, toSet.getModifiers() & ~java.lang.reflect.Modifier.FINAL);
-            findField(name, clazz).set(instance, value);
+            toSet.set(instance, value);
         } catch (NoSuchFieldException | IllegalAccessException e) {
             throw new RuntimeException(e);
         }


rndmorris:
Question I didn't catch before: why are we changing final in the set method, instead of in `findField`, where R is marking fields as accessible?

src/main/java/dev/rndmorris/salisarcana/lib/R.java
```diff
@@ -95,9 +95,9 @@ public <T> T get(String name, Class<T> type) {
     public R set(String name, Object value) {
         try {
             Field toSet = findField(name, clazz);
-            Field modifiersField = findField("modifiers", Field.class);
+            Field modifiersField = findField("modifiers", toSet.getClass());
```
rndmorris, line 98: I'm not sure what this change is accomplishing since `toSet` is *always* going to be a `Field`.

Midnight145: R dev asked me to when I pr'd it upstream and I figured I'd update it here as well

Midnight145:
> Question I didn't catch before: why are we changing final in the set method, instead of in `findField`, where R is marking fields as accessible?

Just to be explicit with usage, if that makes sense. If something is final, it should continue to be final unless we're actively trying to set it. If we're trying to findField, we definitely want the field, so we want to make it accessible. We don't necessarily want to modify the field, so there's no need to make it not-final.

src/main/java/dev/rndmorris/salisarcana/lib/R.java
rndmorris: ... weird... but okie

rndmorris (approved):

PR #198, opened 2025-03-21
Reduce config boilerplate

rndmorris:

Refactoring

Modules and settings all require a bunch of boilerplate to register with their owners.

Modules self-register with the config root, and settings will try to register with the first settings-having setting or module up their dependency tree.

Also deletes now-unused `export-research` command settings.

Also also registers the forgotten `items.Mixin_ItemIconFix` and `items.MixinItemWandRod` with Mixins.java.

No

**Other information**:


rndmorris:
Next up: reorganizing and regrouping settings

src/main/java/dev/rndmorris/salisarcana/config/modules/CommandsModule.java
```diff
@@ -2,73 +2,84 @@
...
+    @Override
+    public void registerSetting(Setting setting) {
+        super.registerSetting(setting);
+        if (setting instanceof CommandSettings cmdSetting && !commandsSettings.contains(cmdSetting)) {
+            commandsSettings.add(cmdSetting);
+        }
+    }
+
+    @Override
+    public void unregisterSetting(Setting setting) {
```
Midnight145, line 78: we don't need the instanceof here, setting will always be instanceof CommandSettings due to the override in registerSetting

rndmorris: Method removed as part of using a `getCommandsSettings` method to filter and return from the inherited settings list instead.

Midnight145 (approved):

PR #204, opened 2025-03-30
Simplified "Single Wand Replacement" mixin in order to avoid a conflict with TC4 Tweaks.

Nikolay-Sitnikov:
Fixes a crash bug with Salis Arcana & TC4 Tweaks.

Minecraft crashes when you open an Arcane Workbench with TC4 Tweaks active.

The mixin is now a tail mixin, which means it does not need to target a method anymore (since TC4 Tweaks completely strips out the entire `onCraftMatrixChanged` function.)

No

**Other information**:


Nikolay-Sitnikov:
According to glee8e, this crash only occurs with old versions of TC4 Tweaks, but this is a good change nonetheless.

rndmorris (approved):

PR #207, opened 2025-03-31
XP shouldn't be automatically sent to a player when a focus is disenchanted

rndmorris:
Tweak

XP is sent to the player as soon as a focus is finished disenchanting

It no longer does that

No

**Other information**:


Midnight145 (approved):

Nikolay-Sitnikov:
I have mentioned the reason why that feature exists: because people who are disenchanting their focus will probably want to re-enchant it as soon as possible, so they need to regain their XP to be able to spend it.

Maybe this could be a config option?

rndmorris:
I disagree that it's useful. It takes two clicks to redeem the refunded XP: one to pick up the item, the other to put it right back in the table.

PR #211, opened 2025-03-31
[bugfix, feat] Right click navigation improvements

Midnight145:
bigfix, feat

buggy behavior when rclick navigation if you close the window with escape

fixes that, also adds botania-style navigation where it saves page on close

no

**Other information**:


rndmorris (changes requested):
Being able to immediately return to the research I had open? Excellent.

**Nitpick:** The player now has to right-click to unwind the entire page stack in order to exit a research item, which is not ideal since right-click is now the only way to return to the research browser.

**Example:** unlock all aspects, then open the aspect research in the 'nomicon. Scroll (I used TC4 tweaks) all the way to the end, then all the way back to the beginning. You then have to right-click many times to exit.

**Suggestion:** If you right-click after navigating further than the first page in a research, perhaps the stack could be automatically unwound until you reach the first page again. A subsequent right-click, without navigating further, could exit the research (as it does now, when the stack is empty).

rndmorris (approved):
New behavior works nicely

PR #212, opened 2025-04-02
various amulet improvements, closes #140

### LINKED ISSUE #140: [Enhancement]: Improve the Amulet of Vis Storage
### Salis Arcana Version

v1.1.1 (latest)

### Context

The Amulet of Vis Storage is practically useless in most contexts, which is an extremely frustrating discovery for anyone who crafts it.
Theoretically, it is really good: it can store 250 vis of each type, and it can recharge just by standing next to a Vis Relay, and it's a bauble.
Practically, it's useless: it has a very slow recharging speed (both input and output), requires you to be much closer to a Vis Relay than you would expect, and (most importantly) it can only recharge your wand **if you're holding it in your hand**.

### Requested Feature

Add configuration options for the range, charging speed from Vis Relays, and allow the wand to charge your entire inventory or hotbar.

### Alternative Solutions

I would be satisfied with just the inventory / hotbar change, since that is what hurts the Amulet of Vis Storage the most. However, the other changes would make the Amulet actually worth wearing because it would allow you to automatically recharge your wand while you're at your base.

### Additional Context

The Thaumonomicon says that these downsides are intentional:

> Your success has been mixed however. You can only recharge wands while holding them in your own hand and even then the recharge rate is slow. 

However, I don't see a reason why these nerfs would be necessary.

Midnight145:
feature

[rndmorris/Salis-Arcana#140](https://github.com/rndmorris/Salis-Arcana/issues/140)

Makes the bounding box size, recharge rate, transfer rate, configurable, and adds a config option to search entire inventory instead of just held item

no

**Other information**:
closes #140

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/config/settings/FloatSetting.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/settings/FloatSetting.java
@@ -1,0 +1,63 @@
+package dev.rndmorris.salisarcana.config.settings;
+
+import net.minecraftforge.common.config.Configuration;
+
+import dev.rndmorris.salisarcana.config.IEnabler;
+
+public class FloatSetting extends Setting {
+
+    private final String name;
+    private final String comment;
+    private final float defaultValue;
+    private float value;
+    private float maxValue = Float.MAX_VALUE;
+    private float minValue = -Float.MAX_VALUE; // Float.MIN_VALUE is the smallest positive value, not the most negative
+
+    public FloatSetting(IEnabler dependency, String name, String comment, float defaultValue) {
+        super(dependency);
+        this.name = name;
+        this.comment = comment;
+        this.defaultValue = defaultValue;
+    }
+
+    @Override
+    public void loadFromConfiguration(Configuration configuration) {
+        value = configuration.getFloat(name, getCategory(), defaultValue, minValue, maxValue, comment);
+    }
+
+    public float getDefaultValue() {
+        return defaultValue;
+    }
+
+    public float getMaxValue() {
+        return maxValue;
+    }
+
+    public float getMinValue() {
+        return minValue;
+    }
+
+    public float getValue() {
+        return value;
+    }
+
+    public float getValueOrDefault() {
+        return enabled ? value : defaultValue;
+    }
+
+    public FloatSetting setMaxValue(float maxValue) {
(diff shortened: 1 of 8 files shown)


Midnight145:
this is all theoretically done, just untested

src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
```diff
@@ -419,6 +419,27 @@ public enum Mixins {
         .addMixinClasses("gui.MixinGuiArcaneWorkbench_MissingResearch", "lib.MixinArcaneSceptreRecipe", "lib.MixinArcaneWandRecipe")
         .addTargetedMod(TargetedMod.THAUMCRAFT)),
 
+    VIS_RELAY_BOX_EXPANSION(new Builder().setPhase(Phase.LATE)
+        .setSide(Side.BOTH)
+        .setApplyIf(ConfigModuleRoot.enhancements.visRelayBoxExpansion::isEnabled)
```
Nikolay-Sitnikov, line 424: How does one disable a `FloatSetting` or `IntSetting`? As far as I know, missing settings are considered enabled with their default values and putting any other value into the config file would enable the setting.

src/main/java/dev/rndmorris/salisarcana/config/settings/FloatSetting.java
```diff
@@ -0,0 +1,63 @@
...
+    public float getMinValue() {
+        return minValue;
+    }
+
+    public float getValue() {
+        return value;
+    }
+
+    public float getValueOrDefault() {
+        return enabled ? value : defaultValue;
```
rndmorris, line 45: Well, this is wrong (and also in IntSetting, which is my fault). `enabled` should be an `isEnabled()` call here.

rndmorris: We actually probably shouldn't have this method on either setting type in the first place, thinking about it.

src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
rndmorris: > As far as I know, missing settings are considered enabled with their default values

This is incorrect. Whether a setting is enabled by default varies from setting to setting.

PR #213, opened 2025-04-03
Fix bugs and exploits around node recharging

rndmorris:
Tweak + bug fix

* Nodes play catch-up based on system time
* Nodes can be drained then quickly unloaded and reloaded to instantly regenerate vis

* Nodes will play catch-up based on world time
* Nodes will track when they were last updated, so they don't play catch-up inappropriately

No

**Other information**:

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/tiles/MixinTileNode_RechargeTime.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/tiles/MixinTileNode_RechargeTime.java
@@ -14,18 +14,19 @@ import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 
 import dev.rndmorris.salisarcana.lib.TimeHelper;
+import dev.rndmorris.salisarcana.lib.ifaces.IGameTimeNode;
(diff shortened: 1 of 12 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/tiles/MixinTileNode_RechargeTime.java
```diff
@@ -0,0 +1,136 @@
...
+import thaumcraft.common.tiles.TileNode;
+
+@Mixin(value = TileNode.class)
+public abstract class MixinTileNode_RechargeTime extends TileThaumcraft implements IGameTimeNode {
+
+    @Shadow(remap = false)
+    public abstract NodeModifier getNodeModifier();
+
+    @Unique
+    long sa$lastTickActive = -1;
```
Midnight145, line 29: Out of curiosity, why measure sa$lastTickActive in ms instead of ticks?

rndmorris: `sa$lastTickActive` *is* measured in ticks. When it gets substituted into the vanilla recharge method, it gets converted to ms to avoid having to rewrite the entire method.

Midnight145 (approved):

PR #220, opened 2025-04-08
Fix crash related to KnowItAll

Nikolay-Sitnikov:
Replaces the `KnowItAll` with a custom class which can be correctly instantiated on the client.

Opening a workbench with an arcane recipe on a multiplayer server crashes the game.
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19522

No longer crashes, and the `KnowItAll` is properly disposed of when the world is closed.

No, except if any PRs are relying on the old location of the `KnowItAll`.

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/lib/KnowItAll.java
+++ b/src/main/java/dev/rndmorris/salisarcana/lib/KnowItAll.java
@@ -34,7 +34,6 @@ public class KnowItAll extends EntityPlayer {
         public void onWorldUnload(WorldEvent.Unload event) {
             if (knowItAll != null && knowItAll.worldObj == event.world) {
                 knowItAll = null;
-                System.out.println("Unloaded KnowItAll");
             }
         }
     }
--- a/src/main/java/dev/rndmorris/salisarcana/lib/KnowItAll.java
+++ b/src/main/java/dev/rndmorris/salisarcana/lib/KnowItAll.java
@@ -1,0 +1,91 @@
+package dev.rndmorris.salisarcana.lib;
+
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.util.ChunkCoordinates;
+import net.minecraft.util.DamageSource;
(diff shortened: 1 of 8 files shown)


Nikolay-Sitnikov:
> Can you PR this to project/main-v2? Main branch is currently "read only" until main-v2 is merged to make the resulting merge conflict as doable as possible.

This is a high-priority bug as far as I can tell - as in, you can't craft anything in the Arcane Workbench until this is fixed. I thought it would be wise to put in into main so the bug doesn't last.

src/main/java/dev/rndmorris/salisarcana/lib/KnowItAll.java
```diff
@@ -0,0 +1,91 @@
...
+        return knowItAll;
+    }
+
+    public static class EventCollector {
+
+        @SubscribeEvent
+        public void onWorldUnload(WorldEvent.Unload event) {
+            if (knowItAll != null && knowItAll.worldObj == event.world) {
+                knowItAll = null;
+                System.out.println("Unloaded KnowItAll");
```
rndmorris, line 37: Is there a reason to use `System.out` instead of `SalisArcana.LOG` here?

rndmorris (commented):
One question. I'd qualify this as an emergency bugfix, and merge conflicts are negligible, so it's fine to go into `main`.

src/main/java/dev/rndmorris/salisarcana/lib/KnowItAll.java
Caedis: Seems to be a left over debug message imo

Nikolay-Sitnikov: Oops. My fault. I'll remove that message in a bit.

Nikolay-Sitnikov: Line has been removed.

rndmorris (approved):

PR #231, opened 2025-04-30
various amulet improvements, closes #140

### LINKED ISSUE #140: [Enhancement]: Improve the Amulet of Vis Storage

Midnight145:
What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)
feature

What is the current behavior? (You can also link to an open issue here)
https://github.com/rndmorris/Salis-Arcana/issues/140

What is the new behavior (if this is a feature change)?
Makes the bounding box size, recharge rate, transfer rate, configurable, and adds a config option to search entire inventory instead of just held item

Does this PR introduce a breaking change?
no

Other information:
closes https://github.com/rndmorris/Salis-Arcana/issues/140


docs/enhancements.md
```diff
@@ -332,3 +332,21 @@ Requires `enableReplaceWandCapsRecipe=true` or `enableReplaceWandCoreRecipe=true
...
+
+# Enhancements - Vis Relay/Amulet Improvements:
+
+## Config Option: `visRelayBoxExpansion`
+
+Increases the range of the vis relay at which items like amulets will be charged. Default is 5, which means players will have to be within 5 blocks of the relay.
+
+## Config Option: `visAmuletRechargeSpeed`
+
+The rate, in ticks, at which the vis amulet will recharge.
```
rndmorris, line 344: The rate at which the amulet will recharge... itself? Or items in your inventory?

src/main/java/dev/rndmorris/salisarcana/config/settings/FloatSetting.java
```diff
@@ -0,0 +1,63 @@
...
+    public float getMinValue() {
+        return minValue;
+    }
+
+    public float getValue() {
+        return value;
+    }
+
+    public float getValueOrDefault() {
+        return enabled ? value : defaultValue;
```
rndmorris, line 45: [From the last version of this PR](https://github.com/rndmorris/Salis-Arcana/pull/212#discussion_r2028792387): We actually probably shouldn't have this method on either setting type (int or float) in the first place. If the value is disabled, `getValue` will return the default anyway. I guess this is leftover from before we wrote an override for `isEnabled`.

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
```diff
@@ -401,6 +402,29 @@ public class ConfigFeatures extends ConfigGroup {
...
+    public final IntSetting visAmuletTickRate = new IntSetting(
+        this,
+        "visAmuletRechargeSpeed",
+        "The rate in ticks at which vis amulets recharge itself.",
+        5).setMinValue(1);
+
+    public final IntSetting visAmuletTransferRate = new IntSetting(
+        this,
+        "visAmuletTransferRate",
+        "The amount of vis an amulet can move at once.",
```
Nikolay-Sitnikov, line 420: Shouldn't this be in units of centivis, rather than vis? Also, how many times per second does the amulet transfer vis?

Also, is this the amount of vis the amulet gains, the amount it gives away per cycle, or the amount it gives away to each item per cycle? (If there is more than one item needing a charge in my inventory, does the amulet drain faster?)

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
```diff
@@ -401,6 +402,29 @@ public class ConfigFeatures extends ConfigGroup {
...
+    public final FloatSetting visRelayBoxExpansion = new FloatSetting(
+        this,
+        "visRelayBoxExpansion",
+        "The amount to expand the bounding box of vis relays. This is used to increase the range at which amulets are charged.",
+        5.0F).setMinValue(0.0F);
+
+    public final IntSetting visAmuletTickRate = new IntSetting(
+        this,
+        "visAmuletRechargeSpeed",
+        "The rate in ticks at which vis amulets recharge itself.",
```
Nikolay-Sitnikov, line 414: ... recharge **themselves**. (small grammar error)

src/main/java/dev/rndmorris/salisarcana/lib/PlayerHelper.java
```diff
@@ -25,4 +30,15 @@ public static int getExperienceForLevel(int level) {
...
+
+    public static Slot[] getItemsInInventory(EntityPlayer player, Class<? extends Item> itemClass) {
+        ArrayList<Slot> slots = new ArrayList<>();
+        for (int i = 0; i < player.inventory.getSizeInventory(); i++) {
+            ItemStack stack = player.inventory.getStackInSlot(i);
+            if (stack != null && itemClass.isInstance(stack.getItem())) {
+                slots.add(player.inventoryContainer.getSlot(i));
+            }
+        }
+        return slots.toArray(new Slot[0]);
```
Nikolay-Sitnikov, line 42: Any reason why the list of slots is converted to an array instead of being returned as a list?

Nikolay-Sitnikov (changes requested):

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
Nikolay-Sitnikov: Also, this text should mention the amount of CV that a Vis Amulet will gain per cycle.

src/main/java/dev/rndmorris/salisarcana/lib/PlayerHelper.java
Midnight145: forgot to push the changes i had where I made it a stream in ada61cfcdebc8bfb3f769bca2ad4d6953ac441a6 but it is intentional

rndmorris (changes requested):
Appears to work fine, except for the noted crash

rndmorris (changes requested):
Some weirdness discovered during testing:

### Charging Speed
Some odd behavior where an empty, non-held wand in my inventory would appear to charge, but then still be completely empty if I updated it (such as if I moved it to a different slot in my inventory).

To reproduce:
1. Set `visAmuletTransferRate` to a very high number (I used 10000)
2. Don a fully-charged vis amulet
3. Craft a new, empty wand, and put it in your inventory. Do not put it in your hand.
4. Vis should appear to instantly transfer from the amulet to the wand
5. Move the wand to a different slot
6. Check the wand and amulet's vis levels and see that the amulet has been drained but the wand is empty


### Charging range
Increasing `visRelayBoxExpansion`, such as to `10.0`, appears to have no effect, and the amulet still only charges at range 5.

rndmorris (approved):
Both issues raised appear to have been resolved in testing.

rndmorris (approved):

rndmorris (approved):

PR #233, opened 2025-04-30
add mob vis drop whitelist, closes #214

### LINKED ISSUE #214: [Enhancement]: Whitelist for mobs dropping vis if possible
### Salis Arcana Version

salisarcana-v1.1.1-main.55+41d6e31fd6-dirty

### Context

Image

Image

### Requested Feature

Whitelist for mobs that can drop vis orbs

### Alternative Solutions

don't let any mobs drop vis orbs?

### Additional Context

我爱你midnight

Midnight145:
feat

#214 

closes #214 

no

**Other information**:

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
@@ -402,16 +402,16 @@ public class ConfigFeatures extends ConfigGroup {
         "If enabled, the potency level of an equal trade or excavation focus will modify its harvest level by one level per level of potency.")
             .setCategory("harvestLevels");
 
-    public final StringArraySetting mobVisWhitelist = new StringArraySetting(
+    public final StringArraySetting mobVisDropList = new StringArraySetting(
         this,
-        "mobVisWhitelist",
-        "Whitelist of entities that can generate vis orbs when killed. If empty, all entities will generate vis orbs.",
+        "mobVisDropBlacklist",
+        "List of entities that can not generate vis orbs when killed.",
         new String[0]);
 
-    public final ToggleSetting mobVisBlacklist = new ToggleSetting(
+    public final ToggleSetting mobVisWhitelist = new ToggleSetting(
         this,
-        "mobVisBlacklist",
-        "If enabled, the mob vis whitelist will be treated as a blacklist instead.").setEnabled(false);
+        "mobVisDropWhitelist",
+        "If enabled, the blacklist will be treated as a whitelist instead instead.").setEnabled(false);
 
     public final ToggleSetting deadlyGazeMobCheck = new ToggleSetting(
         this,
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
@@ -11,6 +11,7 @@ import dev.rndmorris.salisarcana.config.settings.IntArraySetting;
 import dev.rndmorris.salisarcana.config.settings.IntSetting;
 import dev.rndmorris.salisarcana.config.settings.ReplaceWandComponentSettings;
 import dev.rndmorris.salisarcana.config.settings.Setting;
+import dev.rndmorris.salisarcana.config.settings.StringArraySetting;
 import dev.rndmorris.salisarcana.config.settings.ToggleSetting;
 import dev.rndmorris.salisarcana.lib.IntegerHelper;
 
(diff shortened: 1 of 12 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/events/MixinEventHandlerEntity.java
```diff
@@ -0,0 +1,51 @@
...
+
+    }
+
+    @WrapOperation(
+        method = "livingTick(Lnet/minecraftforge/event/entity/living/LivingDeathEvent;)V",
+        at = @At(
+            value = "INVOKE",
+            target = "Lthaumcraft/common/lib/utils/EntityUtils;getRecentlyHit(Lnet/minecraft/entity/EntityLivingBase;)I"))
+    private int sa$shouldGenerateVisOrbs(EntityLivingBase e, Operation<Integer> original) {
+        if (sa$isBlacklist) {
```
Nikolay-Sitnikov, line 40: Can be simplified into
```java
if(sa$isBlacklist != sa$entities.contains(e.getClass())) {
    return original.call(e);
}
return 0;
```

src/main/java/dev/rndmorris/salisarcana/config/settings/StringArraySetting.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public void loadFromConfiguration(Configuration configuration) {
+        this.value = configuration.get(getCategory(), this.name, this.defaultValue, this.comment)
+            .getStringList();
+    }
+
+    public String[] getValue() {
+        return value;
+    }
```
Nikolay-Sitnikov, line 34: Override `isEnabled()` so it either checks if `this.value` contains the same strings as `this.defaultValue` or it reads some form of paired `ToggleSetting` like `IntArraySetting`. Right now, there's no way to disable this setting.

src/main/java/dev/rndmorris/salisarcana/config/settings/StringArraySetting.java
```diff
@@ -0,0 +1,40 @@
...
+import dev.rndmorris.salisarcana.config.IEnabler;
+
+public class StringArraySetting extends Setting {
+
+    protected final String name;
+    protected final String comment;
+    protected final String[] defaultValue;
+    protected String[] value;
+
+    protected final HashSet<Class<? extends Entity>> entities = new HashSet<>();
```
rndmorris, line 17: Query: What is `entities` doing here? It doesn't appear to be used by anything in this class, and looks out of place.

src/main/java/dev/rndmorris/salisarcana/lib/MixinHelpers.java
rndmorris, line 1: Thought: the majority of this class seems to be related to 'nomicon navigation.

Suggestion (non-blocking): this class should be renamed to reflect what it does, and the new method put in another class. (`EntityHelper`, maybe?)

src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/events/MixinEventHandlerEntity.java
rndmorris: It's fine as-is. The explicitness makes the intention behind it very obvious on even an initial skim.

src/main/java/dev/rndmorris/salisarcana/config/settings/StringArraySetting.java
Midnight145: whoops, leftover from when it was gonna be an EntityArraySetting instead before I gave up on that

src/main/java/dev/rndmorris/salisarcana/config/settings/StringArraySetting.java
```diff
@@ -0,0 +1,35 @@
...
+            .getStringList();
+    }
+
+    public String[] getValue() {
+        return value;
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return super.isEnabled() && value != defaultValue;
```
rndmorris, line 33: Concern: as-is, `value != defaultValue` is a reference comparison, not a value comparison. Use `!defaultValue.equals(value)` instead, please.

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/config/settings/StringArraySetting.java
rndmorris: Oof, this is an array reference instead. Ignore my proposed solution.

rndmorris (approved):

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigFeatures.java
```diff
@@ -401,6 +402,17 @@ public class ConfigFeatures extends ConfigGroup {
         "If enabled, the potency level of an equal trade or excavation focus will modify its harvest level by one level per level of potency.")
             .setCategory("harvestLevels");
 
+    public final StringArraySetting mobVisWhitelist = new StringArraySetting(
```
rndmorris, line 405: I have a couple of design nitpicks here.
1. Putting a blank line in this setting in the config will register this setting as activated, which feels like unintended behavior.
2. It feels incongruent to say "this is a whitelist" but also "it has no effect if its empty". It might require a small bit more logic to activate the mixin, but think I would prefer if this was a blacklist by default and the mixin only active if either the list is non-empty (accounting for blanks) or the whitelist/blacklist setting is set to whitelist.

rndmorris (changes requested):

rndmorris (approved):

PR #244, opened 2025-05-07
create an IEntitySelector to better filter entities for deadly gaze potion effect, closes #241

### LINKED ISSUE #241: [Bug]: Deadly Gaze will anger mobs that are invulnerable or immune to Wither
### Salis Arcana Version

latest `main` (commit `a184cce`)

### Where is the bug?

Thaumcraft 4

### Describe the bug

Looking at mobs that cannot be damaged by Deadly Gaze will still anger them. Current list includes:
- Mobs with `{Invulnerable: 1b}` set.
- Wither Skeletons
- Withers / Ender Dragons (although it probably doesn't affect their AI much)
- CustomNPCs that are undamageable or immune to status effects

### To Reproduce

1. Summon a Wither Skeleton
2. Look at it from a distance - observe that it hasn't noticed you
3. Give yourself Deadly Gaze
4. The Wither Skeleton will immediately notice you, even though it wasn't damaged

### Expected Behavior

Deadly Gaze shouldn't anger what it can't damage.
(Also, maybe Deadly Gaze should be a separate effect from Wither? Although it would prevent items like the Witherless Rose from countering its effect...)

### Attachments

_No response_

### Additional Context

Deadly Gaze also won't anger any mobs that are already affected by Wither, even if they would get damaged by Deadly Gaze otherwise. (Whether this is a good thing is up for debate.)

Midnight145:
Bugfix

#241 

fixes that

no

**Other information**:

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/lib/DeadlyGazeEntitySelector.java
+++ b/src/main/java/dev/rndmorris/salisarcana/lib/DeadlyGazeEntitySelector.java
@@ -1,0 +1,31 @@
+package dev.rndmorris.salisarcana.lib;
+
+import net.minecraft.command.IEntitySelector;
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.boss.EntityDragon;
+import net.minecraft.entity.boss.EntityWither;
+import net.minecraft.entity.monster.EntitySkeleton;
+import net.minecraft.potion.Potion;
+import net.minecraft.potion.PotionEffect;
+
+/**
+ * This class is used to filter out entities that should not be affected by the
+ * Deadly Gaze effect.
+ *
+ * @author Midnight145
+ */
+public class DeadlyGazeEntitySelector implements IEntitySelector {
+
+    @Override
+    public boolean isEntityApplicable(Entity entity) {
+        if (entity instanceof EntityLivingBase living) {
+            return living.isEntityInvulnerable()
+                || living.isPotionApplicable(new PotionEffect(Potion.wither.getId(), 1))
+                || (living instanceof EntitySkeleton skeleton && skeleton.getSkeletonType() != 0)
+                || living instanceof EntityWither
+                || living instanceof EntityDragon;
+        }
+        return false;
+    }
+}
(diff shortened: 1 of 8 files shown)


Nikolay-Sitnikov:
Would it be better to make the `IEntitySelector` a singleton rather than instantiating it every tick? (Although it probably doesn't matter in the long run.)

src/main/java/dev/rndmorris/salisarcana/lib/DeadlyGazeEntitySelector.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public class DeadlyGazeEntitySelector implements IEntitySelector {
+
+    @Override
+    public boolean isEntityApplicable(Entity entity) {
+        if (entity instanceof EntityLivingBase living) {
+            return !(living.isEntityInvulnerable()
+                || !living.isPotionApplicable(new PotionEffect(Potion.wither.getId(), 1))
+                || (living instanceof EntitySkeleton skeleton && skeleton.getSkeletonType() != 0)
+                || living instanceof EntityWither
```
rndmorris, line 26: Query: if the goal of checking for a wither or dragon is to exclude bosses, would it make more sense to exclude anything that implements `IBossDisplayData`?

src/main/java/dev/rndmorris/salisarcana/lib/DeadlyGazeEntitySelector.java
```diff
@@ -0,0 +1,31 @@
...
+import net.minecraft.potion.Potion;
+import net.minecraft.potion.PotionEffect;
+
+/**
+ * This class is used to filter out entities that should not be affected by the
+ * Deadly Gaze effect.
+ *
+ * @author Midnight145
+ */
+public class DeadlyGazeEntitySelector implements IEntitySelector {
```
rndmorris, line 18: I believe a static singleton would be reasonable here, since this class has no instance data nor any required setup or cleanup.

src/main/java/dev/rndmorris/salisarcana/lib/DeadlyGazeEntitySelector.java
Midnight145: Wither and Ender Dragon specifically are immune to wither, and while many bosses likely are, it's also likely that many aren't. (probably Blightfall's Tainted Hero or Tainted King, for example, though I'm not positive.)

rndmorris: Works for me, then.

rndmorris (approved):

PR #247, opened 2025-05-10
Re-render any blocks when the biome is changed, closes #245

### LINKED ISSUE #245: [Bug]: Biome changes do not cause chunk rebuilds
### Salis Arcana Version

Latest `main` (commit `bcebd89`)

### Where is the bug?

Thaumcraft 4

### Describe the bug

When the biome of a column is changed (via Taint in a Bottle or Ethereal Blooms), the biome coloration of the column isn't updated until a block change causes the chunk to rebuild the chunk vertices. (Or you can press F3+A to rebuild all chunks.)

### To Reproduce

1. Create a 9x9 carpeted floor high above the ground
2. Splash it with Taint in a Bottle - the carpets will prevent any taint from being placed
3. Observe that the biome coloration of any grass underneath the platform has not changed.
4. Observe that the F3 screen correctly reports the biome as having changed to Tainted Land
5. Rebuild the chunk via F3+A or by placing or breaking a block
6. Observe that the color updates, but only in that chunk

### Expected Behavior

The chunk should rebuild as soon as the biome changes inside it, thus synchronizing the perceived biomes with the actual biomes.

### Attachments

_No response_

### Additional Context

_No response_

Nikolay-Sitnikov:
Adds new bugfix to visually update chunks when biome color is changed.

Closes #245 

Biome colors are updated on the client.

No

### DIFF
--- a/docs/bugfixes.md
+++ b/docs/bugfixes.md
@@ -127,3 +127,9 @@ Use Forge's fishing lists to determine what fish, junk, and treasure a fishing g
 **Config option:** `silverwoodLogCorrectName`
 
 Non-vertical silverwood logs will be correctly named "Silverwood Log" in WAILA.
+
+## Visually Update Biome Colors
+
+**Config option:** `updateBiomeColorRendering`
+
+Biome changes will correctly update the color of grass in chunks without needing a block to change.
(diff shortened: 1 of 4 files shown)


Nikolay-Sitnikov:
This PR has been tested to work in Singleplayer and Multiplayer.

Midnight145 (approved):

PR #248, opened 2025-05-11
Catch parsing errors when parsing enum configs, closes #235

### LINKED ISSUE #235: [Bug]: The Eldritch Altar spawning method config crashes the game if set to invalid value
### Salis Arcana Version

latest `main` (commit `a184cce`)

### Where is the bug?

Salis Arcana

### Describe the bug

`EnumSetting`s don't check to see if the string used in the config is one of the valid strings for that enum, resulting in an `java.lang.IllegalArgumentException: No enum constant` exception.

### To Reproduce

1. Set `S:eldritchAltarSpawningMethod=` in `enhancements.cfg` to an invalid value.
2. Observe that the game will not load with this config set.

### Expected Behavior

The invalid value should be replaced with the default value (with a warning emitted to console.)

### Attachments

_No response_

### Additional Context

_No response_

Nikolay-Sitnikov:
Bug fix - fixes a crash caused by incorrectly set up configurations

Closes #235 

Incorrectly configured enum configs are ignored and the default value is used.

No

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/config/settings/EnumSetting.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/settings/EnumSetting.java
@@ -42,6 +42,11 @@ public class EnumSetting<E extends Enum<E>> extends Setting {
             .map(Enum::toString)
             .toArray(String[]::new);
         final var valueString = configuration.getString(name, getCategory(), value.toString(), comment, validValues);
-        value = Enum.valueOf(enumClass, valueString);
+
+        try {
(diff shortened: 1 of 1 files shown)


Nikolay-Sitnikov:
We could also add logging to invalid configs. However, that would require creating a Logger instance before the mod class is loaded. Loading the mod class early could interfere with other mods trying to ASM our mod's main class. We either need to move the Logger singleton to the `SalisArcanaCore` class or find another method to log errors.

Midnight145 (approved):

PR #249, opened 2025-05-11
Added better filtering & verification to config settings.

Nikolay-Sitnikov:
General improvement - makes sure that int array settings are filtered for invalid values, reports errors which aren't getting silently suppressed by Forge's config system.

`IntArraySetting`s will happily accept out-of-bound numbers (`IntSetting`s & `FloatSetting`s silently clamp the values to the minimum and maximum.)

`IntArraySetting`s can now be required to have a maximum or exact amount of items.
`IntArraySetting`s will now reject out-of-bounds entries, log them, and either reset to default or skip the invalid value.
The Salis Arcana Logger has been moved to `SalisArcanaCore` (with it getting copied in `SalisArcana`) in order to allow core-mods and config processing to read the values without loading `SalisArcana` and potentially disrupting others' ASM.
`EnumSetting`s now log whenever they encounter an invalid value.

No

**Other information**:
`StringSetting` and `HexColorSetting` are unused and have their own host of problems. I don't want to touch them unless I know how they're supposed to function (can they be disabled? should they parse colors properly? should they accept `#FFFFFF` and `0xFFFFFF`?) They should either be removed or improved to follow some sort of convention.

Many potential config mistakes are being silently corrected or ignored by Forge. There's not much we can do about that.

The Eldritch Altar setting's logic is most likely broken. Testing it to confirm is difficult, though, because of the contrived setup. Can I just prove it via code?

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/SalisArcana.java
+++ b/src/main/java/dev/rndmorris/salisarcana/SalisArcana.java
@@ -2,7 +2,6 @@ package dev.rndmorris.salisarcana;
 
 import java.util.Map;
 
-import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import cpw.mods.fml.common.Mod;
@@ -13,6 +12,7 @@ import cpw.mods.fml.common.event.FMLPreInitializationEvent;
 import cpw.mods.fml.common.event.FMLServerStartingEvent;
 import cpw.mods.fml.common.network.NetworkCheckHandler;
 import cpw.mods.fml.relauncher.Side;
+import dev.rndmorris.salisarcana.core.SalisArcanaCore;
 
 @Mod(
     modid = SalisArcana.MODID,
@@ -22,8 +22,8 @@ import cpw.mods.fml.relauncher.Side;
     dependencies = "required-after:Thaumcraft;after:tc4tweak@[1.5.32,)")
 public class SalisArcana {
 
-    public static final String MODID = "salisarcana";
-    public static final Logger LOG = LogManager.getLogger(MODID);
+    public static final String MODID = SalisArcanaCore.MODID;
+    public static final Logger LOG = SalisArcanaCore.LOG;
 
     @SidedProxy(
         clientSide = "dev.rndmorris.salisarcana.ClientProxy",
--- a/src/main/java/dev/rndmorris/salisarcana/SalisArcana.java
+++ b/src/main/java/dev/rndmorris/salisarcana/SalisArcana.java
@@ -2,6 +2,7 @@ package dev.rndmorris.salisarcana;
 
 import java.util.Map;
 
+import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import cpw.mods.fml.common.Mod;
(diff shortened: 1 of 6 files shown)


src/main/java/dev/rndmorris/salisarcana/SalisArcana.java
```diff
@@ -22,8 +22,8 @@
     dependencies = "required-after:Thaumcraft;after:tc4tweak@[1.5.32,)")
 public class SalisArcana {
 
-    public static final String MODID = "salisarcana";
-    public static final Logger LOG = LogManager.getLogger(MODID);
+    public static final String MODID = SalisArcanaCore.MODID;
```
rndmorris, line 25: Personal preference: don't touch `SalisArcanaCore` outside of core-related materials. Just duplicate the string. We're not going to be changing the mod id any time soon.

src/main/java/dev/rndmorris/salisarcana/SalisArcana.java
```diff
@@ -22,8 +22,8 @@
...
+    public static final Logger LOG = SalisArcanaCore.LOG;
```
rndmorris, line 26: I would prefer to have a separate logger instance for core and non-core. Please leave this one as-is with the `salisarcana` modid as the logger's name.

src/main/java/dev/rndmorris/salisarcana/core/SalisArcanaCore.java
```diff
@@ -11,6 +14,9 @@
 @IFMLLoadingPlugin.MCVersion("1.7.10")
 public class SalisArcanaCore implements IFMLLoadingPlugin {
 
+    public static final String MODID = "salisarcana";
+    public static final Logger LOG = LogManager.getLogger(MODID);
```
rndmorris, line 18: As mentioned elsewhere, I'd prefer to have separate loggers for core and non-core uses. In this case, please name this logger `salisarcana-core`. The mod id field here is fine.

src/main/java/dev/rndmorris/salisarcana/config/settings/IntArraySetting.java
```diff
@@ -45,12 +50,60 @@ public boolean isEnabled() {
         return pairedSetting.isEnabled();
     }
 
+    public IntArraySetting setLengthFixed(boolean isFixed) {
+        this.fixedLength = isFixed;
+
+        if (isFixed) {
```
rndmorris, line 56: Personal preference: let each builder method only adjust the thing it's named after. `setFixedLength` can update the `fixedLength` field, but should leave `maxLength` alone.

src/main/java/dev/rndmorris/salisarcana/config/settings/IntArraySetting.java
```diff
@@ -45,12 +50,60 @@ public boolean isEnabled() {
...
+            this.maxLength = this.defaultValue.length;
+        }
+
+        return this;
+    }
+
+    public IntArraySetting setMaxLength(int maxLength) {
+        if (!this.fixedLength) {
```
rndmorris, line 64: Relatedly, let `setMaxLength` do its job without silently depending on `fixedLength`.

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/config/settings/IntArraySetting.java
Nikolay-Sitnikov: The thing is, Forge fixes the length of the array to the maximum length. `maxLength` serves the purpose of "required array length" when `fixedLength` is true.

I'll probably just add a ternary to the `loadFromConfiguration` instead.

rndmorris (approved):

PR #267, opened 2025-05-30
 Prevent eldritch boss doors & nodes being broken by water buckets

Nikolay-Sitnikov:
Bug fix - prevents some blocks from being broken by water buckets

Nodes, energized nodes, and the invisible blocks making up the sides of the Outer Lands boss door can be replaced by water buckets, because water buckets don't check the same breaking conditions as normal breaking. This could allow you to break into the boss room without activating the door first.

An even captures all bucket pick-up and placement events, and cancels them if they're targeting a `blockAiry` that shouldn't be broken. (`FillBucketEvent` is misnamed - it triggers for both filling and emptying buckets.)

No

### DIFF
--- a/docs/bugfixes.md
+++ b/docs/bugfixes.md
@@ -136,7 +136,7 @@ Biome changes will correctly update the color of grass in chunks without needing
 
 ## Prevent Nodes & Boss Doors from being broken by filled Buckets
 
-**Config option:** `airyBlocksNoBucketBreak`
+**Config option:** `preventBlockAiryFluidReplacement`
 
 Prevents useful airy blocks (nodes, energized nodes, and the blocks of the Outer Lands boss room door) from being replaced by buckets with liquid.
 
--- a/src/main/java/dev/rndmorris/salisarcana/CommonProxy.java
+++ b/src/main/java/dev/rndmorris/salisarcana/CommonProxy.java
@@ -83,7 +83,7 @@ public class CommonProxy {
             .register(new StartupNotifications());
         MinecraftForge.EVENT_BUS.register(KnowItAll.EVENT_COLLECTOR);
 
-        if (SalisConfig.bugfixes.airyBlocksNoBucketBreak.isEnabled()) {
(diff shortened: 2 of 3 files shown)


src/main/java/dev/rndmorris/salisarcana/lib/BlockAiryBucketInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+            int z = event.target.blockZ;
+
+            switch (event.target.sideHit) {
+                case 0:
+                    y--;
+                    break;
+                case 1:
+                    y++;
+                    break;
+                case 2:
```
Midnight145, line 25: should 2, 3 be x and 4, 5 be z?

Docs say it goes east/west, north/south, and I know ForgeDirection is x is eas/west, z is north/south.

Also eww, why is event.target.sideHit an int instead of a ForgeDirection. It'd be so much cleaner if it wasn't.

docs/bugfixes.md
```diff
@@ -134,6 +134,12 @@ Non-vertical silverwood logs will be correctly named "Silverwood Log" in WAILA.
 
 Biome changes will correctly update the color of grass in chunks without needing a block to change.
 
+## Prevent Nodes & Boss Doors from being broken by filled Buckets
+
+**Config option:** `airyBlocksNoBucketBreak`
```
Midnight145, line 139: Nonblocking, but I don't love this config name, maybe something like `preventBlockAiryFluidReplacement`?

src/main/java/dev/rndmorris/salisarcana/lib/BlockAiryBucketInterceptor.java
Nikolay-Sitnikov: Nope. Look at `ForgeDirection`'s source - the ±Z directions are 2 & 3. (Also, I practically copied this switch statement from `ItemBucket` with minor modifications.)

Midnight145 (approved):

PR #274, opened 2025-06-04
Implement `getPickBlock` for the Warded Jars, giving the correct item when middle-clicked in Creative Mode

Nikolay-Sitnikov:
Bug fix - improves jar pick-block behavior

Pick-blocking a jar returns just the base block - an empty jar, a Brain in a Jar, or a zero-aspect Node in a Jar.

Pick-block now returns the actual block (as defined by the block dropped when you break the jar.)

No

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/blocks/MixinBlockJar_PickBlock.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/blocks/MixinBlockJar_PickBlock.java
@@ -1,0 +1,33 @@
+package dev.rndmorris.salisarcana.mixins.late.blocks;
+
+import net.minecraft.block.BlockContainer;
+import net.minecraft.block.material.Material;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.MovingObjectPosition;
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/blocks/MixinBlockJar_PickBlock.java
```diff
@@ -0,0 +1,36 @@
...
+
+    protected MixinBlockJar_PickBlock(Material p_i45386_1_) {
+        super(p_i45386_1_);
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public ItemStack getPickBlock(MovingObjectPosition target, World world, int x, int y, int z) {
+        final var drop = this.getDrops(world, x, y, z, world.getBlockMetadata(x, y, z), 0);
+        if (drop.size() == 1) {
```
Midnight145, line 30: q: is there ever a time where `drop.size() > 1`?

Nikolay-Sitnikov: No, but there might be a theoretical branch which could produce `drop.size() == 0`. (If, for some reason, there's no Tile Entity ready for this block, for instance.)

Midnight145 (dismissed):

PR #276, opened 2025-06-05
Make upgraded foci with a different casting vis cost than the base focus spend the correct cost.

### LINKED ISSUE #269: [Bug]: Fix vis cost of upgraded foci
### Salis Arcana Version

2.1.0

### Where is the bug?

Thaumcraft 4

### Describe the bug

Giving wand foci certain upgrades is supposed to also affect their vis cost. For example:
- Chain Lightning adds a small aqua Cost to the Shock Focus
- Fireball adds a small perditio cost to the Fire Focus
- Ice Boulder adds a small terra cost to the Frost Focus

This is reflected in their tooltips, except... it's not real? You can spam as much chain-lighting as you want but the wand never seems to consume aqua vis.

### To Reproduce

1. Upgrade a Fire Focus with Fireball
2. Observe the tooltip now claims a cost of 0.66 ignis vis and 0.33 periditio vis
3. Shoot fireballs
4. Only ignis vis is consumed

### Expected Behavior

1. Upgrade a Fire Focus with Fireball
2. Observe the tooltip now claims a cost of 0.66 ignis vis and 0.33 periditio vis
3. Shoot fireballs
4. Both ignis and perditio vis is consumed

You could of course change the tooltip instead but judging by Azanor's code, the original intention definitely seems to be that the vis cost does in fact change.

### Attachments

_No response_

### Additional Context

_No response_

Nikolay-Sitnikov:
Bug fix - causes upgraded foci to spend the correct vis cost.

Closes #269

Foci will check the focus item in order to determine present upgrades instead of checking the wand item (which cannot have focus upgrade NBT tags.)

No

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/api/MixinItemFocusBasic_WandUpgradeLevel.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/api/MixinItemFocusBasic_WandUpgradeLevel.java
@@ -1,0 +1,24 @@
+package dev.rndmorris.salisarcana.mixins.late.api;
+
+import net.minecraft.item.ItemStack;
+
+import org.spongepowered.asm.mixin.Mixin;
+
+import com.llamalad7.mixinextras.injector.wrapmethod.WrapMethod;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+
+import thaumcraft.api.wands.ItemFocusBasic;
+import thaumcraft.common.items.wands.ItemWandCasting;
+
+@Mixin(ItemFocusBasic.class)
+public class MixinItemFocusBasic_WandUpgradeLevel {
+
+    @WrapMethod(method = "getAppliedUpgrades", remap = false)
+    public short[] unwrapWandItem(ItemStack focusStack, Operation<short[]> original) {
+        if (focusStack.getItem() instanceof ItemWandCasting wand) {
+            focusStack = wand.getFocusItem(focusStack);
+        }
+
+        return original.call(focusStack);
+    }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/api/MixinItemFocusBasic_WandUpgradeLevel.java
```diff
@@ -0,0 +1,24 @@
...
+import thaumcraft.api.wands.ItemFocusBasic;
+import thaumcraft.common.items.wands.ItemWandCasting;
+
+@Mixin(ItemFocusBasic.class)
+public class MixinItemFocusBasic_WandUpgradeLevel {
+
+    @WrapMethod(method = "getAppliedUpgrades", remap = false)
+    public short[] unwrapWandItem(ItemStack focusStack, Operation<short[]> original) {
+        if (focusStack.getItem() instanceof ItemWandCasting wand) {
+            focusStack = wand.getFocusItem(focusStack);
```
Midnight145, line 19: sanity checking myself, focusStack can't be null here, right? Thaum doesn't do any later checks before dereferencing it, so we need to guarantee it can't be.

Nikolay-Sitnikov: Well, the original function will also cause a `NullReferenceException` if you try to pass it a `null`, and this function shouldn't ever get called on a wand without a focus. I can add a sanity check for that, if you wish?

Midnight145: I assumed it wouldn't ever get called if the wand didn't have a focus, but intellij wasn't looking up function calls in Thaum's code as well for some reason, so I couldn't verify myself. Should be good then.

Midnight145 (dismissed):

Midnight145 (approved):

PR #277, opened 2025-06-05
Patterning banners with essentia now only consumes essentia, not the phial

### LINKED ISSUE #172: [Bug]: Banners consume the phial upon being patterned
### Salis Arcana Version

Latest main (commit `3d7e6d6`)

### Where is the bug?

Thaumcraft 4

### Describe the bug

When you use a phial on a banner to put a pattern, not only does it consume the essentia, but it also consumes the Phial that held it. The Phial is also never returned to the user.

### To Reproduce

1. Place a banner
2. Use any phial on it
3. The phial is deleted.

### Expected Behavior

The Phial shouldn't get consumed. The essentia probably also shouldn't get consumed, but that can be an optional setting.

### Attachments

_No response_

### Additional Context

This should be an extremely easy bug to fix, since the cause is only on one line. (Also, isn't there a more correct method of consuming an item than just decrementing the stack size, as what happens currently?)

Nikolay-Sitnikov:
Bug fix - banners no longer consume the phial when being patterned

Closes #172

Banners will now only consume the essentia from the phial when they get patterned by default. The setting can also be changed to make banner patterning completely free.

No


Nikolay-Sitnikov:
Thanks for the tip. I wasn't sure where exactly I needed to write "closes ___", so I wrote it in both the title and the description just to make sure.

src/main/java/dev/rndmorris/salisarcana/config/settings/ConsumePhialSetting.java
```diff
@@ -0,0 +1,27 @@
...
+import dev.rndmorris.salisarcana.config.IEnabler;
+
+public class ConsumePhialSetting extends EnumSetting<ConsumePhialSetting.Option> {
+
+    public ConsumePhialSetting(IEnabler dependency, String name, String comment, @Nonnull Option defaultValue) {
+        super(dependency, name, comment, defaultValue);
+    }
+
+    @Override
+    public boolean isEnabled() {
```
rndmorris, line 14: Move this to EnumSetting and compare `value` against `defaultValue` instead please.

src/main/java/dev/rndmorris/salisarcana/config/settings/ConsumePhialSetting.java
```diff
@@ -0,0 +1,27 @@
+package dev.rndmorris.salisarcana.config.settings;
+
+import javax.annotation.Nonnull;
+
+import dev.rndmorris.salisarcana.config.IEnabler;
+
+public class ConsumePhialSetting extends EnumSetting<ConsumePhialSetting.Option> {
+
+    public ConsumePhialSetting(IEnabler dependency, String name, String comment, @Nonnull Option defaultValue) {
```
rndmorris, line 9: Since this is a very specific class, it's fine to hard-code `defaultValue` to `ESSENTIA` here instead of taking it as a parameter.

rndmorris (changes requested):
Functions as described.

src/main/java/dev/rndmorris/salisarcana/config/settings/ConsumePhialSetting.java
Nikolay-Sitnikov: This function is "is the mixin enabled" - it has a different default than the default of the setting, because the setting is turned on by default.

Nikolay-Sitnikov: In simple terms: if the setting is set to PHIAL, nothing should happen. The setting is not set to PHIAL by default, because by default something should happen. Like a `ToggleSetting`, the `ConsumePhialSetting` is enabled by default. (I could implement this behavior into `EnumSetting`, though)

rndmorris (commented):
One last request/adjustment (that I thought then never actually communicated): this feels a bit more like a tweak/enhancement than an outright bugfix, primarily because of the free option. Thoughts on putting it moving it or breaking it up @Midnight145?

Midnight145:
> One last request/adjustment (that I thought then never actually communicated): this feels a bit more like a tweak/enhancement than an outright bugfix, primarily because of the free option. Thoughts on putting it moving it or breaking it up @Midnight145?

Yeah, I think it's worth keeping the fix in bugfixes and moving the free option to enhancements, and breaking up those mixins

Nikolay-Sitnikov:
I could split the mixins, but there are three different configs targeting this one spot, and getting the mixins to apply in the right order isn't something I want to do.

rndmorris (changes requested):
Pulled the latest changes. Deleted config files for a full refresh. Set `bannerFreePatterns` to false, and set `bannerReturnPhials` to false. Right-clicked a banner in survival mode, neither the phial nor essentia were consumed.

Nikolay-Sitnikov:
I'm pretty dumb to overlook that, fixing now.

PR #281, opened 2025-06-22
Prevent some potential glitches in the Thaumatorium

Nikolay-Sitnikov:
Bug fix - patches the `eventHandler` antipattern in the Thaumatorium

There are a number of bugs in the Thaumatorium, some of which might be caused by this glitch (I'm relatively certain an out-of-bounds exception is possible to trigger using two players interacting with a Thaumatorium at the same time.)

Prevents the container from staying around in memory on the client side and allows the server to properly update the containers of all the players interacting with the Thaumatorium.

No


rndmorris (commented):
Looks fine. I'm assuming it's been tested in both multiplayer and singleplayer to ensure it behaves as expected?

Nikolay-Sitnikov:
Marking as draft for now until I figure out how to prevent the `MultiContainer` from leaking `Container`s.

PR #282, opened 2025-06-22
Localize all chat messages on the client to provide correctly translated text in Multiplayer.

Nikolay-Sitnikov:
Bug fix - makes the "You lack the necessary knowledge to understand this research." message get localized on the client, rather than the server.

Any person playing on a multiplayer server will see this message in English rather than their native language.

Fixes the bug described above.

No


Nikolay-Sitnikov:
Note: This exact mistake is repeated 3 more times in other files. Those will most likely be added to this PR.

Nikolay-Sitnikov:
I checked that the PR runs obfuscated on split client & server, and confirmed that a couple of the mixins actually work.

Unfortunately, confirming that all the mixins work was outside of my abilities. I was unable to trigger warp effects to test their messages, and some of the other mixins also require unique conditions.

"Message gets localized on the server" tests:
- [x] Tapping an Alembic
- [ ] Attempting to bind a mirror to an already linked mirror
- [x] Trying to open the GUI of a Focal Manipulator without the necessary research
- [ ] Enraging Thaumcraft boss with high-damage attacks
- [x] Binding a Hand Mirror
- [x] Using a Hand Mirror after breaking the bound mirror block
- [ ] Using Iron & Gold Keys (partially tested, needs two players to completely test.)
- [ ] Using Research Notes without the prerequisite research
- [ ] Triggering warp events
- [ ] Trying to use flight items in the Outer Lands
- [ ] Research packet without prerequisites, somehow?
- [ ] Opening the Outer Lands boss room door

"Message never got localized" tests: (not completely tested, because no non-English translations are available)
- [ ] Trying to interact with someone else's golem
- [x] Throwing Ender Pearl next to Warded Blocks
- [ ] Changing more of Arcane Pressure Plate
- [ ] Trying to link a Mirror to an already-paired Mirror.

It's probably fine to merge as-is, since all of these messages aren't that important, the methods used across all the mixins are extremely similar, and I really don't have the time to exhaustively test this entire PR.

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemKey_LocalizeCorrectly.java
```diff
@@ -0,0 +1,60 @@
...
+        .setItalic(true);
+
+    @WrapOperation(
+        method = "onItemUseFirst",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/util/StatCollector;translateToLocal(Ljava/lang/String;)Ljava/lang/String;"))
+    public String collectKeys(String langKey, Operation<String> original, @Share("key1") LocalRef<String> keyOne,
+        @Share("key2") LocalRef<String> keyTwo) {
+        if (keyOne.get() == null) {
```
rndmorris, line 35: Some comments explaining _why_ we need `keyOne` and `keyTwo` here would be appreciated since the why behind it is not entirely clear at face value.

From poking around, it looks to be for handling the key3+key4 and key5+key6 pairs?

src/main/java/dev/rndmorris/salisarcana/mixins/late/lib/events/MixinEventHandlerEntity_LocalizeCorrectly.java
```diff
@@ -0,0 +1,85 @@
...
+    public IChatComponent disableFlyComponent(IChatComponent original) {
+        return new ChatComponentTranslation("tc.break.fly").setChatStyle(salisArcana$disableFlyStyle);
+    }
+
+    @ModifyConstant(
+        method = "playerInteract",
+        constant = @Constant(stringValue = "You are not my Master!"),
+        remap = false)
+    public String golemPlayerRejectionKey(String constant) {
+        return "salisarcana:misc.golem.reject_owner";
```
rndmorris, line 53: Nitpick: the golem isn't rejecting its *owner* just a player who isn't its owner, so the key name is a bit inaccurate. `reject_player` would be fine, I think?

rndmorris (changes requested):
Small requests, otherwise looks fine.

Nikolay-Sitnikov:
I want to do a bit more testing, but I have fixed the issues and resolved the merge conflicts.

Nikolay-Sitnikov:
I have confirmed the two-key mixins to also function. Feel free to merge.

Issue #265, opened 2025-05-28
[Bug]: Swapping wand parts costs significantly less than intended


Akeianova:
### Salis Arcana Version

1.1.16-GTNH

### Where is the bug?

Salis Arcana

### Describe the bug

When swapping parts on a wand the vis cost is much lower than the intended cost.

### To Reproduce

- Attempt to craft the wand normally
- Attempt to swap parts on a wand to recreate the same wand
- See the vis cost be less than half the original cost

### Expected Behavior

Switching the parts on the wand should cost as much as creating the wand from scratch

### Attachments

![Image](https://github.com/user-attachments/assets/a36bd657-3c0e-4830-a946-30c9667e5217)
![Image](https://github.com/user-attachments/assets/4fadbfdc-b25b-4a89-8fec-2be74646bf34)

PR #297, opened 2025-07-10
GTNH TC Wands Compat

koolkrafter5:
GTNH TC Wands Compat

https://github.com/rndmorris/Salis-Arcana/issues/265

Screws and conductors are required to upgrade wand cores.
The method from GTNH TC Wands is used to calculate the vis cost instead of base Thaum's.
The Thaumonomicon entries show extra required items and updated vis costs.
Extra items and updated vis costs have relevant configs.
A preexisting bug where staffters could only be upgraded with wand cores and not staff cores has been fixed by adding a STAFFTER enum to WandType.
The thaumcraft-nei-plugin dep is renamed to not cause problems with GTNH TC Wands because it drags in everything on and off the planet transitively.
Functionality without GTW should be unchanged.

I don't think so.

**Other information**:
The logic for the changes to the research inside the lambda is a bit rough and could probably be improved, but I'm not sure how it would be.
Tested as thoroughly as I and @Nikolay-Sitnikov could think to. See conversation in GT:NH Discord: https://discord.com/channels/181078474394566657/603348502637969419/1392856587047014512.

One example:
image

Updated 'nomicon (carrots are filler used by GTW when NH's coremod isn't loaded):
image

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -45,9 +45,21 @@ dependencies {
     compileOnly("maven.modrinth:bugtorch:1.2.13")
     compileOnly(rfg.deobf("maven.modrinth:etfuturum:2.6.2"))
     compileOnly(rfg.deobf("curse.maven:undergroundbiomesconstructs-72744:2304497"))
+    compileOnly("com.github.GTNewHorizons:GTNH-TC-Wands:1.4.5:dev")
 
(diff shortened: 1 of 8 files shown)


rndmorris (approved):
Appears to behave as expected

PR #313, opened 2025-07-18
Add Focus Pouch Slot when Baubles Expanded is installed

### LINKED ISSUE #312: [Enhancement]: Add bauble slot for focus pouch
### Salis Arcana Version

1.1.23

### Context

Too many baubles, not enough slots...

### Requested Feature

Make a dedicated bauble slot for focus pouch/bottomless pouch.

### Alternative Solutions

Uh, make it attachable to the wand/staff somehow?

### Additional Context

_No response_

Nikolay-Sitnikov:
Feature - adds a new slot for the wand
Also fixes several bugs in the old Baubles Expanded mod compat mixin.

Closes #312 

There's a new slot for the Focus Pouch in the Baubles menu

No

**Other information**:
Ideally, the `useAllBaublesSlots` setting should be moved to `mod-compat.cfg`, the related mixin config should require Baubles Expanded to be installed, and the new `focusPouchSlot` setting should have `useAllBaublesSlots` as a prerequisite. If you wish, I can perform those changes.

(Moving the config setting shouldn't be risky, since I doubt anyone ever touched that setting.)


src/main/java/dev/rndmorris/salisarcana/CommonProxy.java
```diff
@@ -63,6 +65,11 @@ public void preInit(FMLPreInitializationEvent event) {
         if (SalisConfig.features.enableFocusDisenchanting.isEnabled()) {
             DisenchantFocusUpgrade.initialize();
         }
+
+        if (Loader.isModLoaded("Baubles|Expanded")) {
```
Midnight145, line 69: request: make an entry in common.compat.Mods for Baubles Expanded, and make a dedicated BaublesExpandedCompat for this and create a preinit method in ModCompat.java (assuming this needs to be in preinit)

Nikolay-Sitnikov: There's no point in making an entry in `common.compat.Mods`, since the `BaseCompatSetting` doesn't integrate with that enum. Also, we now have two enums of mods, `Mods` and `TargetedMod`, both keeping track of the same thing. Should I try to integrate all of this code?

Midnight145 (approved):
Looks good, thanks!

PR #320, opened 2025-07-20
create IVisContainer interface

### LINKED ISSUE #315: [Enhancement]: Add a common interface for Vis Amulets and Wands
### Salis Arcana Version

n/a

### Context

Vis amulets and wands share a number of methods with identical signatures and purposes for managing their vis.

### Requested Feature

Add a shared interface for vis amulets and wands that should (hopefully) make working with them less of a hassle. Put it in the API package for good measure.

### Alternative Solutions

Create the interface, but don't mix it into the items directly. Instead create one or more wrapper classes that take in a wand/vis amulet instance and redirect those method calls to the wrapped object. (I think I prefer this one, honestly).

### Additional Context

Concern: as a general design goal, all mixins should have at least one config option that disables them. If not using the alternative solution: either any logic that uses this interface will have a hard dependency on the mixins being enabled, or you'll end up writing that duplicate code instead anyway.

Submitted on @Midnight145's behalf.

Midnight145:
feat

#315 

closes #315 

no

**Other information**:
As this mixin is unintrusive, and is in the api package for other developers to use, it is flagged as a required mixin.


Midnight145:
Tested

Midnight145:
forgot to push

rndmorris (approved):
Looks fine, no immediate crashes when putting a wand on the recharge pedestal.

PR #321, opened 2025-07-21
Make the upgraded vis cost of Wand Focus: Excavation determined by the upgrade rather than the load order.

Nikolay-Sitnikov:
Bug fix - causes the vis cost of Wand Focus: Excavation to depend solely on the applied upgrade, rather than the order that the upgrades were loaded.

The price of upgraded foci depends on which order the foci upgrades were loaded in.
(these videos were created immediately after the game was restarted.)

https://github.com/user-attachments/assets/c0ae6cf3-57e5-4841-9b30-abfa506e630f

https://github.com/user-attachments/assets/d3146bbd-5777-45e4-bdab-79b2f116bf25

The prices depend only on the upgrade.

No

**Other information**:
There's really no good reason to avoid an `@Overwrite` here. Once you remove all the wrong code and add in the correct code, there's basically no more original code left.


src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemFocusExcavation_DeterministicCost.java
```diff
@@ -0,0 +1,55 @@
...
+    @Shadow
+    @Final
+    private static AspectList cost;
+
+    /**
+     * @author nicksitnikov
+     * @reason The original method poorly implements lazy loading, resulting in nondeterministic results. There's no
+     *         good way to transform the function otherwise.
+     */
+    @Overwrite
```
rndmorris, line 45: > There's really no good reason to avoid an @Overwrite here. Once you remove all the wrong code and add in the correct code, there's basically no more original code left.

Overwrites, to my understanding, cause problems with other mixins that target the same method, which is why we avoid them. If it can be done with something less drastic, we should take the less drastic option. Please replace the overwrite with something less invasive.

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemFocusExcavation_DeterministicCost.java
Nikolay-Sitnikov: They interfere with other mixins if the other mixins target something not present in the overwritten version (or if the priorities are in the wrong order.)

The best I can do is make the issues caused by the mixins targeting some error-filled code silently disappear.

In simple terms - as far as I know, anything that targets `isUpgradedWith`, `cost`, or any return statements will still work. Anything which targets something else has no way to work no matter how I write my mixin, because that code is all wrong.

src/main/java/dev/rndmorris/salisarcana/mixins/late/items/MixinItemFocusExcavation_DeterministicCost.java
```diff
@@ -0,0 +1,59 @@
...
+        .add(Aspect.ORDER, 1)
+        .add(Aspect.ENTROPY, 1)
+        .add(Aspect.EARTH, 16);
+
+    @Unique
+    private static final AspectList sa$dowsingCost = new AspectList().add(Aspect.FIRE, 2)
+        .add(Aspect.ORDER, 2)
+        .add(Aspect.EARTH, 15);
+
+    @ModifyExpressionValue(
```
rndmorris, line 32: An in-code comment explanation of what the cause of the issue is, and why these two mixins fix it, would be appreciated.

rndmorris (commented):
Otherwise it looks fine.

rndmorris (approved):
A lot more in-depth than expected, so good job.

PR #323, opened 2025-07-23
Remove the ability to upgrade a staff using its own vis.

Nikolay-Sitnikov:
Bug fix

Staves can be used to upgrade themselves, thus allowing impossible staves to be created.

Single Wand Replacement doesn't work on Staves & Staffters anymore.

No

**Other information**:
This feels a lot more unintuitive to me. I might look into expanding that "research error" message to show other messages, like "staves cannot supply vis for crafting" & "conductors & screws required to swap wand cores".


rndmorris (commented):
Code change looks correct. I want to test it later today just to be triply sure.

rndmorris (approved):
Quick tests look fine.

PR #351, opened 2025-08-13
Prevent wands from crashing when they have nonexistent components

### LINKED ISSUE #349: [Bug]: Removed Wand Core Crash Loop
### Salis Arcana Version

v2.2.0

### Where is the bug?

Thaumcraft 4

### Describe the bug

If a mod adds a wand core and a player creates a wand using it, the world will crash during loading if that wand core is no longer provided by any loaded mod. This makes the world unplayable until the mod that adds the core is reinstalled or the item is otherwise removed from the save.

### To Reproduce

1. Make a wand with a certain core (for example, the Baguette wand core from Thaumic Exploration)
2. Remove that wand core (disable the config for the Baguette core)
3. Try to load in to the world
4. Observe the crash

### Expected Behavior

The world does not crash. Maybe replace unfindable cores with a stick or have some other kind of fallback logic. It looks like it tries to get the wand core and it returns null.

### Attachments

[Wand_Crash.txt](https://github.com/user-attachments/files/21745490/Wand_Crash.txt)

### Additional Context

Removing cap items probably also has similar issues.

Nikolay-Sitnikov:
Bug fix

Closes #349

Wands without a valid component now use a nonfunctional invalid component.
Wands with invalid amounts of vis no longer overflow the GUI.

No


rndmorris (approved):
Tested, appears to work as expected.

Issue #356, opened 2025-08-18
[Enhancement]: Allow dispensers to shoot primal arrows


koolkrafter5:
### Salis Arcana Version

-

### Context

Dispensers can shoot regular arrows, but not primal arrows

### Requested Feature

Dispensers should be able to shoot primal arrows

### Alternative Solutions

-

PR #357, opened 2025-08-20
Primal arrows can be fired from dispensers

rndmorris:
Tweak

#356 

They can be fired from dispensers

No

**Other information**:

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/common/BehaviorDispensePrimalArrow.java
+++ b/src/main/java/dev/rndmorris/salisarcana/common/BehaviorDispensePrimalArrow.java
@@ -14,9 +14,12 @@ public class BehaviorDispensePrimalArrow implements IBehaviorDispenseItem {
 
     @Override
     public ItemStack dispense(IBlockSource source, ItemStack stack) {
+        // This is implemented similarly to how potions are dispensed
         return new BehaviorProjectileDispense() {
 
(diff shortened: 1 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/common/BehaviorDispensePrimalArrow.java
```diff
@@ -0,0 +1,27 @@
...
+    public ItemStack dispense(IBlockSource source, ItemStack stack) {
+        return new BehaviorProjectileDispense() {
+
+            @Override
+            protected IProjectile getProjectileEntity(World worldIn, IPosition position) {
+                final var entity = new EntityPrimalArrow(worldIn, position.getX(), position.getY(), position.getZ());
+                entity.type = stack.getItemDamage() % 6;
+                return entity;
+            }
+        }.dispense(source, stack);
```
Nikolay-Sitnikov, line 25: I can see that the reason you did this is in order to use a `BehaviorProjectileDispense` without storing (thread-unsafe) state in this singleton. Adding a comment explaining this would be nice, but not required.

Nikolay-Sitnikov (approved):
The code itself is simple & without mistakes. I assume you tested that it works, so I am approving this PR.

Midnight145 (approved):

PR #372, opened 2025-09-14
Expand focus upgrade packet limit

### LINKED ISSUE #369: [Bug]: Focus disenchanting doesn't work
### Salis Arcana Version

Salis-Arcana-1.1.30-GTNH

### Where is the bug?

Salis Arcana

### Describe the bug

Clicking the upgrade button does nothing after putting an enchanted shock focus in the focal manipulator and selecting the disenchant option. (makes clicking sound but button isn't pressed and nothing happens)

### To Reproduce

cheat in clean wand focus: shock, focal-enchant it with potency I, then attempt to disenchant it.

### Expected Behavior

Should do something or have a clearer indication of why it's broken, other foci disenchant also don't seem to work (e.g furgal V maintenance focus)

### Attachments

video in linked discord message: [link](https://discord.com/channels/181078474394566657/522098956491030558/1414382777591402637)

### Additional Context

gtnh version is latest daily before beta4

Nikolay-Sitnikov:
Bug fix - closes #369 

`GuiFocalManipulator` uses a vanilla "enchantment button pressed" packet to send the selected ID of the upgrade to the server. Unfortunately, this packet was not designed to support huge amounts of buttons. While `C11PacketEnchantItem` could store an integer, it would only write a byte of that integer when serialized.

The result was that, when playing on multiplayer servers, focus upgrade IDs greater than 127 would be received as negative numbers on the server & would get ignored.

Instead of sending a vanilla `C11PacketEnchantItem`, we send a custom `MessageExtendedEnchantItem` packet that can store and transmit an entire integer. (Please do note that focus upgrades are limited to shorts, and considering they're stored in an array, picking extra large IDs causes extremely large arrays to get allocated.)

No

### DIFF
--- a/docs/bugfixes.md
+++ b/docs/bugfixes.md
@@ -269,4 +269,4 @@ Prevents the "Vis in Wand" GUI when the wand is held from showing impossible bar
 
 **Config option:** `extendUpgradeFocusPacket`
 
-Use a larger packet for sending the ID of the focus upgrade being selected, thus allowing the use of focus upgrade IDs > 127 on multiplayer servers.
+Use a larger packet for sending the ID of the focus upgrade being selected, allowing the use of focus upgrade IDs > 127 on multiplayer servers.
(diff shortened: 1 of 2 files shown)


Nikolay-Sitnikov:
> > Please do note that focus upgrades are limited to shorts
> 
> If they're limited to shorts elsewhere, why use an integer for transit?

Because we might want to increase the limit to an integer later, especially if we decide a different ID system is necessary. I can make it a short, though.

jordanamr:
A signed short has a max of 32767, I think we will be fine :D

PR #375, opened 2025-09-27
Infusion preview

### LINKED ISSUE #279: [Enhancement]: Warn players when they attempt to start an infusion recipe without sufficient essentia nearby
### Salis Arcana Version

Latest `main` (commit `f3fce8e`)

### Context

People are always frustrated that they have to double-check their essentia jars and calculate out all the necessary phials when doing an infusion recipe, because starting an infusion recipe without enough essentia will waste essentia and cause instability.

### Requested Feature

When you try to start an infusion craft, you will be prevented from starting the craft and a warning will be displayed. There should be a way to override the warning and start the craft anyways.

### Alternative Solutions

Right now, there aren't any other good solutions to the issue. Something like Automagy's Essentia Locus or Thaumic Energistics can make it easier to check your stocks, but you still have to jump between NEI and your world in order to check the essentia.

Since this is a powerful ability, some gating could be required.
- Creating an additional research that's required to unlock this feature is a good idea, since that is also the ideal way to provide information about this feature to our users.
- An additional crafting step could be required, although this would be somewhat complicated to implement (since the NBT would need to transfer from the item to the block to the completed multiblock.)
- Alternatively, we could require something from the wand the player uses to start the recipe - maybe only scepters made out of good materials can perform the check.

### Additional Context

There may be some complications with Essentia Mirrors, Thaumic Energistics' Infusion Provider, and Automagy's Essentia Locus, since these blocks may cause essentia to be double-counted.

sisyphussy:
Adds infusion preview

Adds infusion preview

Doesn't modify any code, just adds a new feature.

**Other information**:


src/main/java/dev/rndmorris/salisarcana/client/infusion/InfusionPreview.java
```diff
@@ -0,0 +1,348 @@
...
+                        0,
+                        771,
+                        1f
+                );
+
+                GL11.glPopMatrix();
+                current++;
+            }
+            GL11.glPopMatrix();
+            String display = missingEssentia ? EnumChatFormatting.RED + "Not enough essentia" : EnumChatFormatting.GOLD + "Right-Click with wand to start";
```
Nikolay-Sitnikov: Needs localization.

Nikolay-Sitnikov (changes requested):
Also, a systematic problem - the client might not have the chunks on the other side of the mirror loaded. It might be worth doing the instability calculations on the server & sending a packet every second or so.

Also, consider using the custom infusion instability logic that Salis Arcana has. (I've mentioned this previously.)

sisyphussy:
> the client might not have the chunks on the other side of the mirror loaded.

Do you think i should check if the entire region of the linked mirror is loaded, and if it isn't, request that information by the server?

> Also, consider using the custom infusion instability logic that Salis Arcana has. (I've mentioned this previously.)

I remember you saying that it was incomplete or something, and it's currently unused so i'll just remove it for the time being and maybe consider adding it later (I'm honestly not sure how i'd properly balance it)

Also it's currently missing a thaumonomicon entry, I thought i could get it done by today but things got in the way so it'll have to wait till tomorrow.

Nikolay-Sitnikov:
> > the client might not have the chunks on the other side of the mirror loaded.
> 
> Do you think i should check if the entire region of the linked mirror is loaded, and if it isn't, request that information by the server?

If you can neatly accomplish it, sure. I'm just worried about the edge cases where the mirror might be loaded but the jars it's using aren't.

> > Also, consider using the custom infusion instability logic that Salis Arcana has. (I've mentioned this previously.)
> 
> I remember you saying that it was incomplete or something, and it's currently unused so i'll just remove it for the time being and maybe consider adding it later (I'm honestly not sure how i'd properly balance it)

> [nicksitnikov] There's also a couple mixins & stuff, so I wouldn't use it blindly just yet

Whoops. I was trying to say "read the code that uses `InfusionMatrixLogic` before you use it because I don't have the time to read through the code & explain it to you", not that the code was incomplete.

DarkShadow44:
May I ask why you closed this? Seems like a useful addition.

sisyphussy:
I didn't feel like adding the rest of the features. Only things missing are thaumonomicon entry and server-side essentia fetching (which, frankly, I don't think is required). 

If someone wants to finish this PR, by all means feel free to do so.

PR #387, opened 2025-10-25
Scan all command, closes #222

### LINKED ISSUE #222: [Enhancement]: Add admin command with player name where it effectively scans all the items in a modpack that have essentia
### Salis Arcana Version

None, I am plaing GTNH 2.7.3, but as TCInventoryScanning is getting merged into here with 2.8.x?2.9.x? I am writing this report here

### Context

I am always frustrated in the modpack I play when I try to find good items for some essentia, but am unable to do so bcs I haven't scanned that good item that has that essentia I want. 

### Requested Feature

In order to fix this, I would like an admin command that, when put executed witha players name, effectively scans all items in a modpack that have essentia for that player and adds them to the list one can see in NEI when looking at sources of a specific essentia.

### Alternative Solutions

Cheating all items a modpack has into huge chests and scanning them by hand might also be an option, but even with just all items in basic MC these days, not to speak of the amount of items in big modpacks like GTNH, that quickly becomes unfeasible. 

### Additional Context

_No response_

Midnight145:
creates salisarcana-scan-all, closes #222 

#222 

creates the command for it

no

**Other information**:


Midnight145:
currently only scans items you have the research for???
also can be expanded to scan entities as well, just haven't done that

PR #388, opened 2025-10-25
Add starry void block

DarkShadow44:
Add the starry void block from TC5. Tried to match the behavior, but since I had to rewrite it it's not perfect, but I think good enough.
No worldgen changes yet, it's only usable from creative.

**Other information**:


DarkShadow44:
Also, keep in mind that while this is ready for review, this should probably need some more testing. Still keep finding some more edge cases where it behaves weirdly.

Issue #392, opened 2025-10-31
[Bug]: Wand Focus: Equal Trade does not place blocks as a player


Nikolay-Sitnikov:
### Salis Arcana Version

Latest, doesn't matter

### Where is the bug?

Thaumcraft 4

### Describe the bug

Leaf blocks placed by Wand Focus: Equal Trade will decay over time, unlike leaves placed by the player. (Reported in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/22300)

### To Reproduce

1. Place some blocks
2. Get a Wand Focus: Equal Trade & attune it to some kind of leaves
3. Use it to replace the blocks with leaves
4. Observe that the leaves decay over time 

### Expected Behavior

Leaves should be persistent even if they're placed by the wand focus.

PR #395, opened 2025-11-02
Fix Leaves decaying if placed by Equal Trade focus

Kynake:
Bug fix

**What is the current behavior?**
Fixes https://github.com/rndmorris/Salis-Arcana/issues/392

No


Kynake:
I'll update this to use `@Accessor`, in case you decide to use it, but feel free to close this in favor of your fix if you think its redundant.

PR #408, opened 2025-11-21
Add remap for MC target

ah-OOG-ah:
Previously this `@At` wasn't remapped, so it missed in production.


Nikolay-Sitnikov:
Why is this a draft PR?

ah-OOG-ah:
Pretty sure this is all of them. Additionally, this fixes https://github.com/Glease/TC4Tweaks/issues/63

PR #428, opened 2025-12-29
Allow golems to cause player kills with their dart launchers.

### LINKED ISSUE #423: [Enhancement]: Make the dart launcher work with the visor.
### Salis Arcana Version

2.3.0

### Context

The dart launcher does not do player kills when combined with a visor. I wish it did.

### Requested Feature

Make it so the dart launcher counts as a player kill when used by a golem wearing a visor.

### Alternative Solutions

Maybe some sort of visual indicator that shows what counts as a player kill.

### Additional Context

_No response_

Nikolay-Sitnikov:
Bug fix - golem visors don't affect the dart launcher

Closes #423

Whenever an entity takes damage, if the source is a dart of a golem equipped with a visor, then the player damage timer is reset.

No

**Other information**:
`EntityUtils.setRecentlyHit` isn't particularly efficient, since it uses uncached reflection. We should probably make a mixin to replace it with an access transformer or a mixin accessor someday.

Testing this PR is difficult, because golems prioritize using their melee over their launcher and they cannot shoot from more than 3 blocks away. I plan to make another PR to allow golems to fire their darts as far as their actual range (since the other code seems to imply that is what was intended.) I did manage to confirm this works in testing by getting a Blaze to drop Blaze Rods when killed by a dart.

### DIFF
--- a/docs/bugfixes.md
+++ b/docs/bugfixes.md
@@ -323,6 +323,12 @@ When right-clicking a Research Note made from Knowledge Fragments, will use a ra
 
 When right-clicking a Research Note made from Knowledge Fragments, will not generate a research note you already have in your inventory.
 
+## Prevent Crucible Dead-Item Duplication Glitch
+
+**Config option:** `crucibleDeadItemDupe`
+
+Prevent multiple Crucibles from melting/crafting the same item entity in the same tick.
+
 ## Golem Visors affect Golem Dart Launchers
 
 **Config option:** `golemVisorAffectDartLauncher`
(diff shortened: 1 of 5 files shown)


rndmorris:
> golems prioritize using their melee over their launcher

Golems respect the barrier created by paving stones of warding, so you should be able to use those to separate the test target from the golem and force it to use its dart launcher.

Nikolay-Sitnikov:
> > golems prioritize using their melee over their launcher
> 
> Golems respect the barrier created by paving stones of warding, so you should be able to use those to separate the test target from the golem and force it to use its dart launcher.

I did that. It just doesn't work well. I don't know what is the problem (maybe it's the distance - 3 blocks is really short, or maybe it's some other check), but they just wouldn't fire consistently.

If you can get it to work smoothly in your testing, more power to you.

PR #432, opened 2026-01-16
Simplify fake-player-lootbag mixin

ah-OOG-ah:
This mixin was failing in GTNH. This version doesn't and should have identical effect, while being much simpler.

It shouldn't, as it should simply make this mixin not fail in prod.


Nikolay-Sitnikov:
> ok thaumic mixins' mixin is... _concerning_. but it should still work as long as this mixin fires later, no?

Huh, I guess so. I didn't realize it used a variable there.

Issue #306, opened 2025-07-16
[Bug]: The breaking animation of the Abandoned Crate causes entities standing on top of the Crate to phase into the Crate


Nikolay-Sitnikov:
### Salis Arcana Version

Latest `main` (commit `991cdff`)

### Where is the bug?

Thaumcraft 4

### Describe the bug

The hitbox of the Abandoned Crate seems to change to a full block while the player is breaking the crate, causing entities to phase into the block and fall through.

### To Reproduce

1. Place down an Abandoned Crate
2. Go into Survival Mode
3. Stand on top of the Abandoned Crate
4. Left-click the Crate once without breaking it.
5. Observe that you have now fallen into the Crate

(The same happens to any other entity on top of the Crate. Item entities, however, will instead bounce because they're programmed to phase *up* through solid blocks.)

### Expected Behavior

Breaking the Abandoned Crate shouldn't change it's hitbox.

### Attachments

https://github.com/user-attachments/assets/ce879cf5-9520-4742-b392-363de362fd4d

Nikolay-Sitnikov:
Apparently, breaking *any* crate or urn will cause all the blocks of the same type to get the wrong hitboxes. Even if you're breaking a crate other than the one you're standing on, you'll still fall through.

PR #433, opened 2026-01-17
Fix all block-bounds alterations that can be fixed.

### LINKED ISSUE #171: [Bug]: Banners Have Broken Collision Boxes
### Salis Arcana Version

Latest main (commit `3d7e6d6`)

### Where is the bug?

Thaumcraft 4

### Describe the bug

Banners from Thaumcraft are supposed to be completely intangible. However, they sometimes have hitboxes that interfere with player movement.

### To Reproduce

1. Place a banner on a wall
2. Jump around next to the banner
3. Eventually, you'll collide with the banner, even though you shouldn't be able to

### Expected Behavior

According to the code, wall banners are supposed to be completely intangible, so you shouldn't be able to collide with them.

### Attachments

_No response_

### Additional Context

I don't know exactly why the bug happens, but a possible cause is that both `setBlockBoundsBasedOnState` and `addCollisionBoxesToList` mutate the `BlockWoodenDevice`'s bounds. (They both call `this.setBlockBounds`)

(Also, how does the `setBlockBoundsBasedOnState` even work? There's only one instance of the `Block` and there are multiple states, so it swaps the `Block`'s data to a different state every time a hitbox is required?)

Nikolay-Sitnikov:
Numerous bug fixes.

Closes #171
Supersedes the fix for #306
Fixes glitchiness when standing on Essentia Reservoirs on a remote server.

All of those bugs are gone and the hitbox behavior of all affected blocks is constant.
This may also allow us to mark these renderers as thread-safe for Angelica in the future.

Refactors our class transformer system.

Potentially. It removes certain methods from certain classes, which could break mixins. I have tested it with TC4 Tweaks and several add-ons, but have not tested it with Thaumic Mixins yet.

**Other information**:
This does not fix the mouse-over behavior of blocks. That will be done in a separate PR.


rndmorris (commented):
I didn't see anything that concerned me. We might need to wait until @Midnight145 has time to look over this, since I believe they're much more familiar with ASM than I am.

Midnight145 (approved):
Looks good, though I'm not nearly familiar enough with the tree api to be able to spot anything besides major issues

Only issue I potentially see:
Is  `org.spongepowered.asm.lib.ClassWriter` safe, or will it load classes?

PR #440, opened 2026-01-23
Collapse name, comment into Setting class instead of every subclass

Midnight145:
cleanup

the Setting class doesn't have a name or comment field, so every subclass adds their own

The Setting class now has those fields and all other classes use those

no

**Other information**:
n/a


Nikolay-Sitnikov (dismissed):
All the code changes look good and clean.

src/main/java/dev/rndmorris/salisarcana/config/settings/HexColorSetting.java
rndmorris: This class isn't in use anywhere, and I don't foresee it making a return, so we can probably scrap it. If we do need it again at some point, we can dig it out of the repo's history.

PR #444, opened 2026-02-17
Ability to Disable Aspect Tint

Cleptomania:
Rendering tweak

Thaumcraft aspect icons are all grayscale and tinted dynamically when rendered.

Some resource packs that want to add icons may provide fully colored icons. This provides an option to disable the tint by hard-setting the aspect colors to white during the rendering.

No, there is a config to control the setting, and it is false by default.

**Other information**:
I'm not sure if this really belongs in salis arcana, this was mostly made at the request of people doing texture work for GTNH. I don't really know much about thaumcraft or what place makes sense for this, so if this mixin/tweak is better suited somewhere else, that's fine.

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinUtilsFX_DisableAspectTint.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinUtilsFX_DisableAspectTint.java
@@ -2,18 +2,20 @@ package dev.rndmorris.salisarcana.mixins.late.thaumcraft.client.lib;
 
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
-import org.spongepowered.asm.mixin.injection.Redirect;
+
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 
 import thaumcraft.api.aspects.Aspect;
 import thaumcraft.client.lib.UtilsFX;
 
 @Mixin(value = UtilsFX.class, remap = false)
-public class MixinUtilsFX_DisableAspectTint {
+public abstract class MixinUtilsFX_DisableAspectTint {
 
-    @Redirect(
+    @WrapOperation(
         method = "drawTag(DDLthaumcraft/api/aspects/Aspect;FIDIFZ)V",
         at = @At(value = "INVOKE", target = "Lthaumcraft/api/aspects/Aspect;getColor()I"))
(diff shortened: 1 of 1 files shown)


june-dev-username:
thanks so much clepto! this works great on my end with the resourcepack. thanks for adding compatibility for this!

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinUtilsFX_DisableAspectTint.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import thaumcraft.api.aspects.Aspect;
+import thaumcraft.client.lib.UtilsFX;
+
+@Mixin(value = UtilsFX.class, remap = false)
+public class MixinUtilsFX_DisableAspectTint {
```
rndmorris: Standardization: Please add the abstract keyword to the class declaration.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -152,6 +152,10 @@ public boolean anyPotionIdOverrideActive() {
         "hiddenResearchUseWorldRandom",
         "When right-clicking a Research Note made from Knowledge Fragments, will use a random value not dependent on the world's time of day.");
 
+    public final ToggleSetting disableAspectTint = new ToggleSetting(this, "disableAspectTint", """
+        When enabled, disables the tinting of aspects images, this can be useful in combination with resource packs
```
rndmorris: Please add this description to `docs/thaumcraft_configuration.md` as well. Top of the file, below `Config file: ...` and above `## Node Behaviors` would be fine.

rndmorris (changes requested):

rndmorris (commented):
Apart from the minor things I've mentioned, this looks good. Thanks for the work!

Cleptomania:
Addressed all the review changes(and re-tested mixin to make sure it was still working the same with switching to WrapOperation)

Nikolay-Sitnikov:
Perhaps we should make this controlled by a setting inside a resource pack? It would make installing resource packs that much easier.

Cleptomania:
I'm not opposed to the idea, but I'm not aware of any mechanism that currently exists for doing that

Nikolay-Sitnikov:
> I'm not opposed to the idea, but I'm not aware of any mechanism that currently exists for doing that

I think I know how to do it, but I will add it in a separate PR.

PR #469, opened 2026-03-23
Add NEI handlers for wand core and cap substitution with Aspect Recipe Index

koolkrafter5:
Feature

There is no NEI integration for wand component substitution.

Two new NEI handlers, one for replacing wand caps and one for replacing wand cores.
Finding uses of wand/staff cores shows a substitution recipe with an example wand or staff.
Finding uses of wand caps shows substitution recipes with an example wand and scepter.
Finding uses of a wand shows all valid (e.g. not trying to put gold caps on a wand that already has them) substitution recipes for researched caps and cores. Respects the enforceWandCoreTypes config setting.
Showing all recipes for either handler shows all possible recipes with example wands.

The core substitution handler supports the extra ingredients required by GTNHTCWands. Both handlers support its different vis cost automatically.

No

**Other information**:
Video of the handler in action:

https://github.com/user-attachments/assets/ad9c3795-cff8-40e9-8e37-95cd83ab3a8f

GTNHTCWands support example:
image

I also replaced TCNEIPlugin in the dev environment with ARI because it reimplements all of its features. If something is missing, let me know.

I also also removed the commented bauble dependency replacer for GTNHTCWands because it was no longer needed.


src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
```diff
@@ -0,0 +1,46 @@
...
+    public static void IMCSender() {
+        if (SalisConfig.features.replaceWandCoreSettings.isEnabled()
+            && SalisConfig.modCompat.apsectrecipeindex.coreReplacementNEIHandler.isEnabled()) {
+            registerHandlerInfo(new WandCoreSubstitutionHandler().getOverlayIdentifier(), "Thaumcraft:WandRod", 144);
+            registerCatalystInfo(new WandCoreSubstitutionHandler().getOverlayIdentifier(), "Thaumcraft:blockTable:15");
+        }
+        if (SalisConfig.features.replaceWandCapsSettings.isEnabled()
+            && SalisConfig.modCompat.apsectrecipeindex.capReplacementNEIHandler.isEnabled()) {
+            registerHandlerInfo(new WandCapSubstitutionHandler().getOverlayIdentifier(), "Thaumcraft:WandCap:1", 144);
+            registerCatalystInfo(new WandCapSubstitutionHandler().getOverlayIdentifier(), "Thaumcraft:blockTable:15");
```
Nikolay-Sitnikov: Technically, wand swaps can be performed in anything that can perform arcane recipes. Is there a way to borrow Aspect Recipe Index's catalyst info, so we list the Arcane Assembler as a valid catalyst if Thaumic Energistics is installed?

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
```diff
@@ -0,0 +1,46 @@
...
+import dev.rndmorris.salisarcana.SalisArcana;
+import dev.rndmorris.salisarcana.config.SalisConfig;
+
+public class IMCForNEI {
+
+    @Optional.Method(modid = "NotEnoughItems")
+    public static void IMCSender() {
+        if (SalisConfig.features.replaceWandCoreSettings.isEnabled()
+            && SalisConfig.modCompat.apsectrecipeindex.coreReplacementNEIHandler.isEnabled()) {
+            registerHandlerInfo(new WandCoreSubstitutionHandler().getOverlayIdentifier(), "Thaumcraft:WandRod", 144);
```
Nikolay-Sitnikov: Can you please move the overlay identifiers into static final fields, so we don't have to construct an object just to read a constant?

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigModCompat.java
```diff
@@ -14,6 +15,7 @@ public class ConfigModCompat extends ConfigGroup {
 
     public final UBCCompatSettings undergroundBiomes = new UBCCompatSettings(this);
     public final GTNHTCWandsCompatSettings gtnhWands = new GTNHTCWandsCompatSettings(this);
+    public final AspectRecipeIndexCompatSettings apsectrecipeindex = new AspectRecipeIndexCompatSettings(this);
```
Nikolay-Sitnikov: `apsectrecipeindex` is misspelled and doesn't follow our naming conventions. It should be `aspectRecipeIndex`.

src/main/java/dev/rndmorris/salisarcana/common/compat/NEIConfig.java
```diff
@@ -18,6 +25,25 @@ public class NEIConfig implements IConfigureNEI {
...
+                API.registerUsageHandler(new WandCoreSubstitutionHandler());
+                API.registerGuiOverlay(GuiArcaneWorkbench.class, "salisarcana.substitution.core", positioner);
+                API.registerGuiOverlayHandler(GuiArcaneWorkbench.class, handler, "salisarcana.substitution.core");
+            }
+            if (SalisConfig.features.replaceWandCapsSettings.isEnabled()
+                && SalisConfig.modCompat.apsectrecipeindex.capReplacementNEIHandler.isEnabled()) {
+                API.registerRecipeHandler(new WandCapSubstitutionHandler());
+                API.registerUsageHandler(new WandCapSubstitutionHandler());
+                API.registerGuiOverlay(GuiArcaneWorkbench.class, "salisarcana.substitution.caps", positioner);
+                API.registerGuiOverlayHandler(GuiArcaneWorkbench.class, handler, "salisarcana.substitution.caps");
```
Nikolay-Sitnikov: Make sure to link these hardcoded strings to the handler's overlay identifier once you move the identifier into a constant.

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
koolkrafter5: ThE will need to add the catalysts itself. It also needs to register its overlay handler for the arcane terminal to work with the recipes.

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
koolkrafter5: I've hated that I needed to make new objects for this and this is a great solution.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigModCompat.java
koolkrafter5: It's a good match for `salisaracan`. Fixed https://github.com/rndmorris/Salis-Arcana/pull/469/commits/799912e832cdabfcf8390cd16170a5c727e432fc

src/main/java/dev/rndmorris/salisarcana/common/compat/NEIConfig.java
koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/469/commits/d41fd72243bc503910076df8ee3de522626a59d4

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/469/commits/d41fd72243bc503910076df8ee3de522626a59d4

koolkrafter5:
> Hmm I guess by making a new wand to avoid needing to deal with the vis in the current one I also make it look like it voids the equipped focus and other things like being renamed in an anvil. I'll fix that tomorrow.

I decided the simplest option was to remove the aspect tags from the new wand so I wouldn't have to see if you could use the wand to swap its own components and try to calculate the remaining vis or try to cap it to the new wand's maximum:
image

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
```diff
@@ -0,0 +1,46 @@
...
+import dev.rndmorris.salisarcana.SalisArcana;
+import dev.rndmorris.salisarcana.config.SalisConfig;
+
+public class IMCForNEI {
+
+    @Optional.Method(modid = "NotEnoughItems")
+    public static void IMCSender() {
+        if (SalisConfig.features.replaceWandCoreSettings.isEnabled()
+            && SalisConfig.modCompat.aspectRecipeIndex.coreReplacementNEIHandler.isEnabled()) {
+            registerHandlerInfo(WandCoreSubstitutionHandler.OVERLAY, "Thaumcraft:WandRod", 144);
```
Nikolay-Sitnikov, line 16: Does this work for add-on wand rods & caps? `Thaumcraft:WandRod` only covers the vanilla rod items.

Nikolay-Sitnikov: If it doesn't, you could probably loop over the registry for Thaumcraft's rods & caps (see `WandHelper` for examples.)

koolkrafter5: That's just to set the item icon shown on the tab, in this case greatwood rod for the rod substitution handler and gold cap for cap substitution handler:

image

Nikolay-Sitnikov:
> I decided the simplest option was to remove the aspect tags from the new wand so I wouldn't have to see if you could use the wand to swap its own components and try to calculate the remaining vis or try to cap it to the new wand's maximum.

I think it would be better to just preserve all the vis in the wand, as if you swapped the caps using another wand. Plus, this way, you could use the player's wand within the recipe directly, as to not confuse the player.

koolkrafter5:
Handling it properly was simpler than I expected since I realized I could just reuse the code from the recipe:

image
image

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/IMCForNEI.java
```diff
@@ -0,0 +1,46 @@
...
+        NBT.setString("modId", SalisArcana.MODID);
+        NBT.setBoolean("modRequired", true);
+        NBT.setString("itemName", stack);
+        NBT.setInteger("handlerHeight", height);
+        FMLInterModComms.sendMessage("NotEnoughItems", "registerHandlerInfo", NBT);
+    }
+
+    @Optional.Method(modid = "NotEnoughItems")
+    private static void registerCatalystInfo(String handlerName, String stack) {
+        NBTTagCompound aNBT = new NBTTagCompound();
```
Nikolay-Sitnikov: Both variables should just be named `nbt`.

koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/469/commits/f90eb5f2df5cb010dd3e0a559d415228417b16f4

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCoreSubstitutionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+import dev.rndmorris.salisarcana.common.compat.GTNHTCWandsCompat;
+import dev.rndmorris.salisarcana.common.recipes.ReplaceWandCoreRecipe;
+import dev.rndmorris.salisarcana.config.SalisConfig;
+import dev.rndmorris.salisarcana.lib.WandType;
+import thaumcraft.api.wands.StaffRod;
+import thaumcraft.api.wands.WandRod;
+import thaumcraft.common.config.ConfigItems;
+import thaumcraft.common.items.wands.ItemWandCasting;
+
+public class WandCoreSubstitutionHandler extends WandCapSubstitutionHandler {
```
Nikolay-Sitnikov: Extending a completely unrelated class just to get access to some constants is a code smell to me. Declare those constants as public or package-private, make `WandCoreSubstitutionHandler` extend `ShapelessArcaneRecipeHandler`, and maybe move the constants to `WandHelper` if you think it is justified and easy.

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCoreSubstitutionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private void generateAllCoreSubstitutionRecipes(ItemStack wandItem, ItemWandCasting wand) {
+        WandType type = WandType.getWandType(wandItem);
+        boolean scepter = type == WandType.SCEPTER || type == WandType.STAFFTER;
+        for (WandRod rod : WandRod.rods.values()) {
+            if (!WandRecipeHandler.validResearch(rod.getResearch()) || rod == wand.getRod(wandItem)
+                || (SalisConfig.features.enforceWandCoreTypes.isEnabled() && !type.isCoreSuitable(rod))) continue;
+            ItemStack outputWand = replaceCore(wandItem, rod);
+            boolean shouldShowRecipe = Util.shouldShowRecipe("salisarcana:REPLACEWANDCORE")
```
Nikolay-Sitnikov: `Util.shouldShowRecipe("salisarcana:REPLACEWANDCORE")` could & should be hoisted out of the loop.

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCapSubstitutionHandler.java
```diff
@@ -0,0 +1,173 @@
...
+        return cap == ConfigItems.WAND_CAP_IRON ? GOLD_GREATWOOD_SCEPTER : IRON_STICK_SCEPTER;
+    }
+
+    private void generateAllCapSubstitutionRecipes(ItemStack wandItem, ItemWandCasting wand) {
+        WandType type = WandType.getWandType(wandItem);
+        boolean scepter = type == WandType.SCEPTER || type == WandType.STAFFTER;
+        for (WandCap cap : WandCap.caps.values()) {
+            if (!WandRecipeHandler.validResearch(cap.getResearch()) || cap == wand.getCap(wandItem)) continue;
+            ItemStack outputWand = replaceCap(wandItem, cap);
+            boolean shouldShowRecipe = Util.shouldShowRecipe("salisarcana:REPLACEWANDCAPS")
```
Nikolay-Sitnikov: Same here.

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCoreSubstitutionHandler.java
Nikolay-Sitnikov: `!scepter || Util.shouldShowRecipe("SCEPTRE")` could be hoisted up & out of the loop too.

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCoreSubstitutionHandler.java
koolkrafter5: Fair enough, they shared more initially but they ended up pretty different as I developed it. https://github.com/rndmorris/Salis-Arcana/pull/469/commits/4a7664c5fe127fa69d7d70836df93b96a2fe524e

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCoreSubstitutionHandler.java
koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/469/commits/15176c8d76f05c973997337ad19c922df41bb759

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCapSubstitutionHandler.java
koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/469/commits/15176c8d76f05c973997337ad19c922df41bb759

src/main/java/dev/rndmorris/salisarcana/common/compat/nei/WandCoreSubstitutionHandler.java
```diff
@@ -0,0 +1,169 @@
...
+import dev.rndmorris.salisarcana.lib.WandType;
+import thaumcraft.api.wands.StaffRod;
+import thaumcraft.api.wands.WandRod;
+import thaumcraft.common.config.ConfigItems;
+import thaumcraft.common.items.wands.ItemWandCasting;
+
+public class WandCoreSubstitutionHandler extends ShapelessArcaneRecipeHandler {
+
+    public static final String OVERLAY = "salisarcana.substitution.core";
+
```
Nikolay-Sitnikov, line 40: This handler also needs to suppress the crafting recipe requests.

```suggestion

    @Override
    public void loadCraftingRecipes(String outputId, Object... results) {
        if (outputId.equals("item")) return;
        loadUsageRecipes(outputId, results);
    }

    @Override
    public void loadCraftingRecipes(ItemStack result) {
        // NO OP
    }
```

Nikolay-Sitnikov:
Waiting for the recipes shown when looking at usages of a wand core to be fixed to include both wands & scepters.

Nikolay-Sitnikov:
Unfortunately, this doesn't quite work as intended. With wand-to-staff swaps permitted, NEI displays two copies of each of the two normal recipes when looking at the usages for staff & wand cores, and doesn't show any recipes that actually switch wand-to-staff or reverse.

koolkrafter5:
Fixed it with the config off:
image
image
image
image

Still works as expected with the config on:
image
image

Nikolay-Sitnikov (dismissed):
Approved, congratulations!

Nikolay-Sitnikov (approved):

PR #473, opened 2026-03-31
Add a null check to fix a crash during Outer Lands generation

koolkrafter5:
Bug fix

Sometimes the Outer Lands generation causes this crash (happened to me in Blightfall, may be made worse by some other mod like archaicfix which is in the stacktrace because I haven't noticed this problem before):
https://mclo.gs/qph8Fc3

It will not attempt to write to a null TE.

No

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
@@ -325,6 +325,11 @@ public class ConfigBugfixes extends ConfigGroup {
         "fixInventoryAspects",
         "Fixes a bug where some GUIs would render the wrong aspects when shifting over a slot. Also improves the performance of this overlay.");
 
+    public final ToggleSetting fixOuterLandsNull = new ToggleSetting(
+        this,
+        "fixOuterLandsNull",
+        "Fixes generation of the Outer Lands dungeons failing due to attempting to set the orientation of an invalid tile entity.");
+
     @Nonnull
     @Override
     public String getGroupName() {
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/world/dim/MixinGenCommon_NullCheck.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/world/dim/MixinGenCommon_NullCheck.java
@@ -1,0 +1,31 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.common.lib.world.dim;
+
+import net.minecraft.world.World;
+import net.sf.cglib.asm.Opcodes;
+
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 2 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/world/dim/MixinGenCommon_NullCheck.java
```diff
@@ -0,0 +1,31 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import com.llamalad7.mixinextras.sugar.Local;
+
+import thaumcraft.common.lib.world.dim.GenCommon;
+import thaumcraft.common.tiles.TileCrystal;
+
+@Mixin(value = GenCommon.class, remap = false)
+public class MixinGenCommon_NullCheck {
```
rndmorris: Standardization: mixin classes should be abstract

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/world/dim/MixinGenCommon_NullCheck.java
koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/473/commits/f36f004ac0484319073e28f76f9a8968f3ebb932

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
```diff
@@ -325,6 +325,11 @@ public class ConfigBugfixes extends ConfigGroup {
         "fixInventoryAspects",
         "Fixes a bug where some GUIs would render the wrong aspects when shifting over a slot. Also improves the performance of this overlay.");
 
+    public final ToggleSetting fixOuterLandsNull = new ToggleSetting(
+        this,
+        "fixOuterLandsNull",
+        "Fixes generation of the Outer Lands dungeons failing due to attempting to set the orientation of an invalid tile entity.");
```
Nikolay-Sitnikov, line 331: Needs an entry in the documentation.

Nikolay-Sitnikov (dismissed):

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
koolkrafter5: https://github.com/rndmorris/Salis-Arcana/pull/473/commits/2e2b25a92f6b9556efa06c4a51bdf296f1ead438

koolkrafter5:
Should be fixed as part of https://github.com/GTNewHorizons/Hodgepodge/pull/822 so this isn't needed

PR #474, opened 2026-04-05
Add a mechanic to the crucible tracking failed goo spills with consequences

thehoblit:

This introduces a new toggleable mechanic to the crucible. It tracks when the crucible fails to generate tainted goo and gas and increases a pressure counter that leads to consequences such as converting the surrounding biome to the tainted biome, increasing nearby players temporary warp, and eventually exploding.

Without this enabled, the player can block off the sides of the crucible to fill it as much as they want with no downsides and craft extremely fast.

The crucible tracks pressure, at low pressure the biome nearby is converted to tainted biome and temporary warp is applied to nearby players. This can happen even if the crucible is completely uncovered but is constantly spitting out goo and gas. If it is covered for an extended amount of time, it will explode with the explosion size scaling based on how full the crucible was. The explosion will also rapidly decay every element, technically increasing the "pressure" under the hood, but as it gets destroyed it is all applied as tainted biome and temporary warp at once. This was not intended at the start, but I kept it in as a feature of  the explosion meltdown.

No.

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -174,6 +174,12 @@ public class ConfigThaumcraft extends ConfigGroup {
         "Improves the particle engine of Thaumcraft by removing unnecessary GL operations.")
             .setCategory(tc4PerformanceCategory);
 
+    public final ToggleSetting cruciblePressure = new ToggleSetting(
+        this,
+        "cruciblePressure",
+        "If true, the Thaumcraft crucible will track when it cannot create flux goo and gas, leading to detrimental and destructive effects.")
+            .setEnabled(false);
+
     @Override
     public @NotNull String getGroupName() {
         return "thaumcraft_configuration";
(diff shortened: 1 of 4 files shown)


koolkrafter5:
Cleaning up flux with cobble or a water bucket is not fun or engaging gameplay, it's just an annoying chore. I do not miss having to do it after learning about blocking the crucible's sides.

thehoblit:
With respect, that is why the Thaumatorium is in the mod, I don't believe mass crafting in the crucible is supposed to be free, easy, and faster. It is supposed to be the early primitive method that you move past once you need to start crafting more of it's recipes.

koolkrafter5:
This punishes all crafting with the crucible. Needing to turn items into essentia to craft something in a Thaumatorium is leagues more annoying than just being able to throw the items you need into a crucible and not needing to worry about the flux that appears when you empty it with your wand.

LazyFleshWasTaken:
While i think this would do well in base thaumcraft, GTNH Thaum is another beast entirely.

There is too much crafting, too much essentia smelting, and the machiens are too slow to keep pace with the crafting.

As suhc, this should either be off for GTNH, or the capacity of the crucible before even flux spills be doubled, or a higher tier crucible be added.

thehoblit:
I agree that it is more work to craft things with the thaumatorium. I just also believe that the crucible should not be so easy and free to use. Mainly the flux spawn blocking making it just absurd to quickly craft stack upon stacks of things.

thehoblit:
> While i think this would do well in base thaumcraft, GTNH Thaum is another beast entirely.
> 
> There is too much crafting, too much essentia smelting, and the machiens are too slow to keep pace with the crafting.
> 
> As suhc, this should either be off for GTNH, or the capacity of the crucible before even flux spills be doubled, or a higher tier crucible be added.

I am fully open to more balance suggestions. I don't have any attachment to the numbers or thresholds  I have employed

PR #477, opened 2026-04-16
Make the crucible's decay scale more with how full it is

thehoblit:

Feature

The vanilla crucible decays 1 aspect every 5 seconds, with a possibility and bias toward breaking it down in to a component. Additionally, it will remove 4 aspects per second when the crucible has more than 100 total aspects. This decay will never break down a compound aspect and will just remove them.

Starting at 200 total aspects, the crucible now decays a percent of the total aspects, with bias toward breaking each aspect down in to a component. This percent slowly increases to 4.2% at 1000 total aspects, 10x the first cap. In practice, it is not always a clean 4.2% as the additional decay will break down compound aspects in to a component, making that "decay" a net 0 aspect change, removing one and adding one.

The starting point of 200 total aspects was chosen to give the default Thaumcraft's decay a place.

The cap of 4.2% was chosen because that is the maximum decay rate vanilla Thaumcraft uses, at exactly 100 aspects it will decay 4 aspects per second and 1 every 5 seconds that breaks down compound aspects.

No.

**Other information**:

This is an iteration on my previous crucible PR after feedback in the GTNH discord: https://github.com/rndmorris/Salis-Arcana/pull/474

### DIFF
--- a/docs/thaumcraft_configuration.md
+++ b/docs/thaumcraft_configuration.md
@@ -6,6 +6,9 @@ Config file: `config/salisarcana/thaumcraft_configuration.cfg`
 If true, disables the tinting of aspect images, this can be useful in combination with resource packs
 that provide images that are already colored.
 
+### Config Option: `crucibleScalingAspectDecay` (Default `true`)
+If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.
+
 ## Node Behaviors (Group `node_behaviors`)
 
 ### Config Option: `hungryDynamicReach` (Default `false`)
--- a/docs/thaumcraft_configuration.md
+++ b/docs/thaumcraft_configuration.md
@@ -7,7 +7,35 @@ If true, disables the tinting of aspect images, this can be useful in combinatio
 that provide images that are already colored.
 
 ### Config Option: `crucibleScalingAspectDecay` (Default `true`)
-If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.
+If true, the crucible's contained aspects will decay at an increasing rate based on how over full the crucible is.
+
+All overfull decay will break down compound aspects in to one of their components.
+
+For example, with the default start of `200`, range of `800`, and maximum rate of `4.2`%, a crucible with 200 total
+aspects will start to experience decay, and the decay will increase for 800 aspects until it reaches a maximum of 4.2%
+at 1000 total aspects and above. Because compound aspects' components are returned to the pool, you will not see a
+strict 4.2% reduction in total aspect count each second.
+
+The vanilla crucible will decay 1 aspect every 5 seconds, splitting any non-primal aspect in to a component.
+When the crucible is over 100 total aspects, it will remove 4 random aspects per second.
+
+### Config Option: `crucibleAspectDecayStart` (Default `200`)
+Requires `crucibleScalingAspectDecay`.
+
+The total aspect count of a crucible at which point the percent scaled decay will start.
+
+### Config Option: `crucibleAspectDecayRange` (Default `800`)
+Requires `crucibleScalingAspectDecay`.
+
+The range, in aspect count, between the point scaling aspect decay starts and the point it reaches its maximum rate.
+
+### Config Option: `crucibleAspectDecayMaximumRate` (Default `4.2`)
+Requires `crucibleScalingAspectDecay`.
+
+The maximum percentage of the crucible's total aspect count that will be removed per second, when the crucible is at or
+above the decay's start plus range.
+
+The default maximum rate of 4.2% is vanilla Thaumcraft's maximum decay rate with its flat decay at 100 total aspects.
 
 ## Node Behaviors (Group `node_behaviors`)
 
--- a/docs/thaumcraft_configuration.md
+++ b/docs/thaumcraft_configuration.md
@@ -7,16 +7,16 @@ If true, disables the tinting of aspect images, this can be useful in combinatio
 that provide images that are already colored.
 
 ### Config Option: `crucibleScalingAspectDecay` (Default `true`)
-If true, the crucible's contained aspects will decay at an increasing rate based on how over full the crucible is.
+If true, the crucible's contained aspects will decay at an increasing rate based on how overfull the crucible is.
 
-All overfull decay will break down compound aspects in to one of their components.
+All overfull decay will break down compound aspects into one of their components.
 
 For example, with the default start of `200`, range of `800`, and maximum rate of `4.2`%, a crucible with 200 total
 aspects will start to experience decay, and the decay will increase for 800 aspects until it reaches a maximum of 4.2%
 at 1000 total aspects and above. Because compound aspects' components are returned to the pool, you will not see a
 strict 4.2% reduction in total aspect count each second.
 
-The vanilla crucible will decay 1 aspect every 5 seconds, splitting any non-primal aspect in to a component.
+The vanilla crucible will decay 1 aspect every 5 seconds, splitting any non-primal aspect into a component.
 When the crucible is over 100 total aspects, it will remove 4 random aspects per second.
 
 ### Config Option: `crucibleAspectDecayStart` (Default `200`)
@@ -33,7 +33,7 @@ The range, in aspect count, between the point scaling aspect decay starts and th
 Requires `crucibleScalingAspectDecay`.
 
 The maximum percentage of the crucible's total aspect count that will be removed per second, when the crucible is at or
-above the decay's start plus range.
+above the decay's maximum range.
 
 The default maximum rate of 4.2% is vanilla Thaumcraft's maximum decay rate with its flat decay at 100 total aspects.
 
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -177,7 +177,8 @@ public class ConfigThaumcraft extends ConfigGroup {
     public final ToggleSetting crucibleScalingAspectDecay = new ToggleSetting(
         this,
         "crucibleScalingAspectDecay",
-        "If true, the Thaumcraft crucible's contained aspects will decay at an increasing rate based on how full the crucible is.");
+        "If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.")
+            .setEnabled(true);
 
     @Override
     public @NotNull String getGroupName() {
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -174,6 +174,11 @@ public class ConfigThaumcraft extends ConfigGroup {
         "Improves the particle engine of Thaumcraft by removing unnecessary GL operations.")
             .setCategory(tc4PerformanceCategory);
 
+    public final ToggleSetting crucibleScalingAspectDecay = new ToggleSetting(
+        this,
+        "crucibleScalingAspectDecay",
+        "If true, the Thaumcraft crucible's contained aspects will decay at an increasing rate based on how full the crucible is.");
+
     @Override
     public @NotNull String getGroupName() {
         return "thaumcraft_configuration";
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -177,8 +177,7 @@ public class ConfigThaumcraft extends ConfigGroup {
     public final ToggleSetting crucibleScalingAspectDecay = new ToggleSetting(
         this,
         "crucibleScalingAspectDecay",
-        "If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.")
-            .setEnabled(true);
+        "If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.");
 
     @Override
     public @NotNull String getGroupName() {
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -179,14 +179,14 @@ public class ConfigThaumcraft extends ConfigGroup {
         this,
         "crucibleScalingAspectDecay",
         """
-            If true, the crucible's contained aspects will decay at an increasing rate based on how over full the crucible is.
-            All over full decay will break down compound aspects in to one of their components.
+            If true, the crucible's contained aspects will decay at an increasing rate based on how overfull the crucible is.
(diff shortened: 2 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -174,6 +174,12 @@ public boolean anyPotionIdOverrideActive() {
         "Improves the particle engine of Thaumcraft by removing unnecessary GL operations.")
             .setCategory(tc4PerformanceCategory);
 
+    public final ToggleSetting crucibleScalingAspectDecay = new ToggleSetting(
+        this,
+        "crucibleScalingAspectDecay",
+        "If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.")
+            .setEnabled(true);
```
rndmorris: This `setEnabled` is redundant; `ToggleSetting` instances are enabled by default.

rndmorris (commented):
Looking at this when I can. Only a nitpick so far.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
thehoblit: Thanks, please nitpick anything you like.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -174,6 +174,11 @@ public boolean anyPotionIdOverrideActive() {
         "Improves the particle engine of Thaumcraft by removing unnecessary GL operations.")
             .setCategory(tc4PerformanceCategory);
 
+    public final ToggleSetting crucibleScalingAspectDecay = new ToggleSetting(
+        this,
+        "crucibleScalingAspectDecay",
+        "If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.");
+
```
Nikolay-Sitnikov, line 204: Needs to be added to the documentation `.md` files.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -0,0 +1,67 @@
...
+            if (!a.isPrimal()) {
+                if (worldObj.rand.nextBoolean()) {
+                    aspects.add(a.getComponents()[0], 1);
+                } else {
+                    aspects.add(a.getComponents()[1], 1);
+                }
+            } else {
+                spill();
+            }
+        }
```
Nikolay-Sitnikov: This has the potential to get quite laggy if someone dumps in a bunch of items, or especially if someone loads a world with an existing billion-essentia Crucible. At the very least, maybe limit the number of times `spill` can get called over the entire method? If you want to do something more advanced, try removing the loop by using binomial distributions - it's fine if only one aspect experiences the rapid decay.

thehoblit: I did both, although not a binomial distribution. I've limited the spills to 5 per cycle, and instead of looping over every individual aspect, I now divide the aspects to be removed over the aspect list and loop over the list to remove in bulk.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
thehoblit: Thanks, I didn't even see those before, I added it to thaumcraft_configuration.md, that seemed like the right place.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -0,0 +1,89 @@
...
+        if (aspects.size() == 0 || removeCount <= 1) return;
+        Random rand = worldObj.rand;
+
+        int splitRemoval = (int) Math.ceil((double) removeCount / aspects.size());
+
+        int toRemove;
+        int remainder = 0;
+        int totalRemoved = 0;
+        while (totalRemoved < removeCount) {
+            Aspect[] availableAspects = aspects.getAspects();
```
rndmorris: `getAspects()` here initializes a new array every time it's called. While probably not a massive deal, doing that just to keep track of which aspects are available every cycle of the loop feels a bit heavy handed.

My suggested alternative would be this:

Before the loop:
* Make a copy of the underlying map  
`final var newAspects = new HashMap<>(aspects.aspects)`
* Make a list of that map's keys  
`final var availableAspects = new ArrayList<>(newAspects.keySet())`


During the loop:
* Operate on `newAspects` instead of `this.aspects`,
* Add/remove values from `availableAspects` manually as aspects rise above or fall to `0`


After the loop:
* Clear the crucible's underlying map  
`aspects.aspects.clear()`
* Then copy your working map's values into the `AspectList`'s  
`aspects.aspects.putAll(newAspects)`


I'm open to feedback on why this suggestion may not be as useful.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -0,0 +1,89 @@
...
+            Aspect a = availableAspects[rand.nextInt(availableAspects.length)];
+            if (a.isPrimal()) {
+                a = availableAspects[rand.nextInt(availableAspects.length)];
+            }
+            toRemove = Math.min(splitRemoval + remainder, removeCount - totalRemoved);
+            if (toRemove > aspects.getAmount(a)) {
```
rndmorris: Suggestion: You might consider replacing this if/else with some `Math.min` and `Math.max` calls:
```java
int aspectAmount = ...;
remainder = Math.max(0, toRemove - aspectAmount);
toRemove = Math.min(toRemove, aspectAmount);
```

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
thehoblit: This is a much better way to handle it, thanks! Brought the instantiations down to a constant 2 instead of sometimes 10+

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
thehoblit: much cleaner thanks

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -0,0 +1,104 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import thaumcraft.api.TileThaumcraft;
+import thaumcraft.api.aspects.Aspect;
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.common.tiles.TileCrucible;
+
+@Mixin(value = TileCrucible.class)
```
rndmorris: During testing I noticed the compiler was emitting warnings that the `@Shadow`ed fields below couldn't be remapped. They all either need `@Shadow(remap = false)` or you need `@Mixin(..., remap = false)` and `@Inject(..., remap = true)`.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Shadow
+    public abstract void spill();
+
+    @Inject(method = "updateEntity", at = @At("TAIL"))
+    private void additionalScaledAspectDrain(CallbackInfo ci) {
+        if (worldObj.isRemote || heat <= 150 || ((int) counter + 1) % 20 != 0) return;
+
+        int total = tagAmount();
+        int excess = total - 200;
```
rndmorris: 200 personally feels kind of low, but it's an alright default. This constant `200` is a good candidate for moving to an entry in `ConfigThaumcraft`. Maybe named `crucibleAspectDecayStart`, with a dependency on `crucibleScalingAspectDecay`?

rndmorris (changes requested):
Otherwise I think this looks good.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
thehoblit: In doing that it made the increasing decay rate stuff a bit awkward, so I just made everything configurable. I also added some more info on the mechanic to the markdown.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -25,22 +26,32 @@ public abstract class MixinTileCrucible_ScalingAspectDecay extends TileThaumcraf
     public AspectList aspects;
     @Shadow
     private long counter;
+    @Unique
+    private final int salis_Arcana$aspectDecayStart = SalisConfig.thaum.crucibleAspectDecayStart.getValue();
```
rndmorris: Standardization: the field prefix should be `salisArcana$`.

rndmorris: Concern: I'm not thrilled at the idea of copying these values into every `TileCrucible` instance, and only slightly less dissatisfied at the idea of these being static fields instead.

Opinion: Since the `getValue()` methods on the setting objects being accessed here are standard getters, it should be fine if you called them directly from your methods (assigning to local variables if used more than once).

docs/thaumcraft_configuration.md
```diff
@@ -7,7 +7,35 @@ If true, disables the tinting of aspect images, this can be useful in combinatio
...
+### Config Option: `crucibleAspectDecayRange` (Default `800`)
+Requires `crucibleScalingAspectDecay`.
+
+The range, in aspect count, between the point scaling aspect decay starts and the point it reaches its maximum rate.
+
+### Config Option: `crucibleAspectDecayMaximumRate` (Default `4.2`)
+Requires `crucibleScalingAspectDecay`.
+
+The maximum percentage of the crucible's total aspect count that will be removed per second, when the crucible is at or
+above the decay's start plus range.
```
rndmorris: Suggestion: "above the decay's maximum range" might read better here.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -177,7 +178,29 @@ public boolean anyPotionIdOverrideActive() {
     public final ToggleSetting crucibleScalingAspectDecay = new ToggleSetting(
         this,
         "crucibleScalingAspectDecay",
-        "If true, the crucible's contained aspects will decay at an increasing rate based on how full the crucible is.");
+        """
+            If true, the crucible's contained aspects will decay at an increasing rate based on how over full the crucible is.
+            All over full decay will break down compound aspects in to one of their components.
```
rndmorris: Typo: "in to" should be "into"

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -177,7 +178,29 @@ public boolean anyPotionIdOverrideActive() {
...
+            """);
+
+    public final IntSetting crucibleAspectDecayStart = new IntSetting(
+        crucibleScalingAspectDecay,
+        "crucibleAspectDecayStart",
+        "The total aspect count of a crucible at which point the percent scaled decay will start.",
```
rndmorris: Suggestion (Opinion): "scaling aspect decay" as you have it in the setting below reads better than "percent scaled decay" as you have it here.

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -177,7 +178,29 @@ public boolean anyPotionIdOverrideActive() {
...
+    public final IntSetting crucibleAspectDecayRange = new IntSetting(
+        crucibleScalingAspectDecay,
+        "crucibleAspectDecayRange",
+        "The range, in aspect count, between the point scaling aspect decay starts and the point it reaches its maximum rate.",
+        800).setMinValue(0);
+
+    public final FloatSetting crucibleAspectDecayMaximumRate = new FloatSetting(
+        crucibleScalingAspectDecay,
+        "crucibleAspectDecayMaximumRate",
+        "The maximum percentage of the crucible's total aspect count that will be removed per second at the end of its range.",
```
rndmorris: Suggestion (non-blocking): "at the end of its range" might be unnecessary here.

rndmorris (changes requested):

thehoblit:
Made all changes, turns out it's overfull, not over full too lmao.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileCrucible_ScalingAspectDecay.java
```diff
@@ -66,7 +61,7 @@ private void additionalScaledAspectDrain(CallbackInfo ci) {
     }
 
     @Unique
-    private void salis_Arcana$removeAndSplit(int removeCount) {
+    private void salisArcana$removeAndSplit(int removeCount) {
```
rndmorris: It should actually be `salisarcana$` (no capital A). That's my mistake.

rndmorris (changes requested):
One last thing (mea culpa) then I think we're good.

Thank you for your patience!

thehoblit:
No worries, thanks for all the quality reviewing.

rndmorris (approved):
Looks fine now, behaves as expected.

PR #478, opened 2026-04-26
[Memory-opti:fix leak] fix Thaumcraft's particle engine leaking the world instance

Alexdoru:

Fix particle engine leaking world instance

image

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/fx/MixinParticleEngine_FixLeak.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/fx/MixinParticleEngine_FixLeak.java
@@ -1,0 +1,29 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.client.fx;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import net.minecraft.client.particle.EntityFX;
+import net.minecraftforge.event.world.WorldEvent;
+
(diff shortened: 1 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/fx/MixinParticleEngine_FixLeak.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Shadow
+    private HashMap<Integer, ArrayList<EntityFX>>[] particles;
+
+    @SubscribeEvent
+    public void onWorldUnload(WorldEvent.Unload event) {
+        if (event.world.isRemote) {
+            for (HashMap<Integer, ArrayList<EntityFX>> map : this.particles) {
+                map.clear();
+            }
```
Nikolay-Sitnikov, line 26: You might want to first check whether the world of the particles is not the same as the world currently being rendered. It is possible for a dimension to get unloaded while the player is in another dimension (say, if an entity passes through a Nether Portal while the player remains on the other side in Singleplayer), and when that world gets unloaded, all existing Thaumcraft particles will disappear.

Alexdoru: It checks for world is remote, so it won't fire in the case you are describing

Nikolay-Sitnikov (approved):

Alexdoru:
any news on this ?

PR #480, opened 2026-04-30
[Memory-opti:fix leak] Fix 3 world client leaks in thaumcraft's TESR

Alexdoru:


src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTESR_FixLeak.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Mixin({ TileEldritchCapRenderer.class, TileFocalManipulatorRenderer.class, TileThaumatoriumRenderer.class })
+public abstract class MixinTESR_FixLeak extends TileEntitySpecialRenderer {
+
+    @Shadow(remap = false)
+    EntityItem entityitem;
+
+    @Override
+    public void func_147496_a(World p_147496_1_) {
+        this.entityitem = null;
```
Nikolay-Sitnikov, line 22: This works for when you switch dimensions, but I don't think this will help with unloading the last world once a player "saves & quits to title".

Alexdoru: You are correct in vanilla minecraft this method is not fired when exiting to main menu however in hodgepodge we added a call to this method when exiting to main menu with world = null as argument.

Nikolay-Sitnikov: Alright then, this looks good.

Nikolay-Sitnikov (approved):

PR #487, opened 2026-05-20
Fix stabilizer strength check

DarkShadow44:

This should have stabilizer rewrite working even if stabilizerStrength has its default value

IntSetting has the following override:
```
 @Override
    public boolean isEnabled() {
        return value != defaultValue && super.isEnabled();
    }
```

So unless stabilizerStrength is different from its default value it will be treated as disabled. Which seems wrong, since this skips all the addon logic, tool.

This aims to fix that behavior and allow rewrite even with default strength.

No

**Other information**:

Not sure why that IntSetting override exists, maybe that is the real problem? It got added in c518dc89 (#163).


rndmorris (approved):
Looks good to me, thanks for fixing that.

PR #489, opened 2026-05-25
Fix node rendering transparency layering

thehoblit:

Fixes the transparency rendering bug on nodes, energized nodes, node aspect tags, and node draining beams.

Currently, they render behind transparent blocks such as water and ice when they are in front of them.

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17918

Nope!

**Other information**:

Gadomancy replaces the node renderer with it's own, and will require it's own fix. This fix works for TC4 as it is.


Nikolay-Sitnikov:
Have you checked TC4 Tweaks? It's uses ASM to inject instead of mixins, for some reason.

thehoblit:
> Have you checked TC4 Tweaks? It's uses ASM to inject instead of mixins, for some reason.

I believe the only part of node rendering that TC4Tweaks touches isn't affecting what I am doing here. I am by no means an expert in java asm and have never used or looked at it before.

https://github.com/Glease/TC4Tweaks/blob/master/src/main/java/net/glease/tc4tweak/asm/NodeLikeRendererVisitor.java

They visit the renderFacingStrip function inside the renderNode function, which I am not mixing in to, I just call it from a different place. This works for energized nodes just fine.

Honestly, if  there weren't other mixins touching the base TileNodeRenderer.renderTileEntityAt, I would think that thaums TileNode got connected to a different renderer or something.

Despite the mixins showing up in the class post load, the te is functioning as though they never mixed in and the old functionality is still there. Logging in the mixin shows that it is never even being called  for regular nodes.

thehoblit:
Me just typing that made me want to double check that other mixins are actually applying, and they aren't. At least in the GTNH version I am testing. I did a dummy test with the backhand mixin and it's not running. I will do more testing tomorow.

https://github.com/GTNewHorizons/Backhand/blob/master/src/main/java/xonin/backhand/mixins/late/thaumcraft/MixinTileNodeRenderer.java

thehoblit:
I have found that the issue is Gadomancy replacing the regular node's TE and renderer with it's own. Interestingly, they seemingly got permission from Azanor to modify the render code, at least that is what  the comment leads me to believe.

https://github.com/GTNewHorizons/Gadomancy/blob/50fe31ad2cf2e1c747feb391f143bd45262773ca/src/main/java/makeo/gadomancy/client/renderers/tile/RenderTileNodeBasic.java#L34

Either that or they just straight up stole Azanor's code and he never found out.

On one hand this means that my fixes work for non-Gadomancy packs, but a fix has to be added to Gadomancy too for any pack that uses it.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTileNodeRenderer_FixNodeRendering.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import dev.rndmorris.salisarcana.lib.NodeRenderingQueue;
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.api.nodes.NodeModifier;
+import thaumcraft.api.nodes.NodeType;
+import thaumcraft.client.renderers.tile.TileNodeRenderer;
+
+@Mixin(value = TileNodeRenderer.class)
+public class MixinTileNodeRenderer_FixNodeRendering {
```
rndmorris:
```suggestion
abstract class MixinTileNodeRenderer_FixNodeRendering {
```

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTileNodeRenderer_FixNodeRendering.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Redirect(
```
rndmorris: Standardization: `@WrapMethod` is preferred to `@Redirect`. This will need updated in most of your mixin classes, it looks like.

src/main/java/dev/rndmorris/salisarcana/client/lib/NodeRenderingQueue.java
rndmorris, line 1: Since this is only intended to be used client-side, it should probably be moved to a new `dev.rndmorris.salisarcana.client.lib` package.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinRenderEventHandler_FixNodeRendering.java
```diff
@@ -0,0 +1,34 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import dev.rndmorris.salisarcana.lib.NodeRenderingQueue;
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.client.lib.RenderEventHandler;
+
+@Mixin(value = RenderEventHandler.class, remap = false)
+public class MixinRenderEventHandler_FixNodeRendering {
```
rndmorris:
```suggestion
abstract class MixinRenderEventHandler_FixNodeRendering {
```

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTileEnergizedNodeRenderer_FixNodeRendering.java
```diff
@@ -0,0 +1,40 @@
...
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import dev.rndmorris.salisarcana.lib.NodeRenderingQueue;
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.api.nodes.NodeModifier;
+import thaumcraft.api.nodes.NodeType;
+import thaumcraft.client.renderers.tile.TileNodeEnergizedRenderer;
+
+@Mixin(value = TileNodeEnergizedRenderer.class)
+public class MixinTileEnergizedNodeRenderer_FixNodeRendering {
```
rndmorris:
```suggestion
abstract class MixinTileEnergizedNodeRenderer_FixNodeRendering {
```

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTileNodeRenderer_FixNodeRendering.java
```diff
@@ -0,0 +1,39 @@
...
+        method = "renderTileEntityAt",
+        at = @At(
+            value = "INVOKE",
+            target = "Lthaumcraft/client/renderers/tile/TileNodeRenderer;renderNode(Lnet/minecraft/entity/EntityLivingBase;DZZFIIIFLthaumcraft/api/aspects/AspectList;Lthaumcraft/api/nodes/NodeType;Lthaumcraft/api/nodes/NodeModifier;)V"))
+    private void redirectRenderNode(EntityLivingBase viewer, double viewDistance, boolean visible, boolean depthIgnore,
+        float size, int x, int y, int z, float partialTicks, AspectList aspects, NodeType type, NodeModifier mod) {
+        NodeRenderingQueue.nodeQueue.add(
```
rndmorris: Readability (non-blocking): it might be cleaner to put the `nodeQueue.add(...)` call behind a `NodeRenderingQueue.queueNode(...)` method. Same for the other queue types.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTileEnergizedNodeRenderer_FixNodeRendering.java
```diff
@@ -0,0 +1,40 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import dev.rndmorris.salisarcana.lib.NodeRenderingQueue;
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.api.nodes.NodeModifier;
+import thaumcraft.api.nodes.NodeType;
+import thaumcraft.client.renderers.tile.TileNodeEnergizedRenderer;
+
+@Mixin(value = TileNodeEnergizedRenderer.class)
```
rndmorris, line 18:
```suggestion
@Mixin(value = TileNodeEnergizedRenderer.class, remap = false)
```

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/renderers/tile/MixinTileNodeRenderer_FixNodeRendering.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import dev.rndmorris.salisarcana.lib.NodeRenderingQueue;
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.api.nodes.NodeModifier;
+import thaumcraft.api.nodes.NodeType;
+import thaumcraft.client.renderers.tile.TileNodeRenderer;
+
+@Mixin(value = TileNodeRenderer.class)
```
rndmorris, line 17:
```suggestion
@Mixin(value = TileNodeRenderer.class, remap = false)
```

src/main/java/dev/rndmorris/salisarcana/lib/NodeRenderingQueue.java
```diff
@@ -0,0 +1,119 @@
...
+import thaumcraft.api.nodes.NodeType;
+import thaumcraft.client.lib.UtilsFX;
+import thaumcraft.client.renderers.tile.TileNodeRenderer;
+
+public class NodeRenderingQueue {
+
+    public static final List<QueuedNode> nodeQueue = new ArrayList<>();
+    public static final List<QueuedLightning> lightningQueue = new ArrayList<>();
+    public static final List<QueuedDrainBeam> drainQueue = new ArrayList<>();
+    public static final List<Runnable> tagQueue = new ArrayList<>();
```
rndmorris: Query: why `Runnable` here instead of stashing `RenderEventHandler instance` inside a queue entry record?

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/lib/NodeRenderingQueue.java
thehoblit: Honestly I just saw that it wasn't static and made it that way, didn't even think to just use  the instance. It looks much cleaner with using the record queue though, thanks.

thehoblit:
I didn't quite follow your suggestion with the remapping, the targeted methods remap but the wrapped methods inside do not, so it's awkward to get it to play nice with the dev compiler and in a pack.

thehoblit:
I've learned more about rendering and I think I've found a more elegant way to fix this

thehoblit:
So my  initial attempts haven't really worked. and I have more or less given up on fighting Minecraft's rendering for proper  transparency layering.

What my original queue fix did was really push all the node rendering to the front so that it _always_ rendered on top of transparent blocks. This looked like it worked properly because 99% of the time you are looking at anything node related, its dead on with nothing between you and the node. Without goggles of revealing making nodes always visible, the node is so small you will never really notice it rendering behind things it shouldn't. It does also leave energized nodes on top, but I was fine with that as they are, well, _energized_.

Overriding the shouldRenderInPass to return pass == 1 will make the nodes render on top of everything transparent on it's own. The only thing the queue is useful for is the hover aspect tags, which are not attached to any tile entity.

In my ventures to fix other rendering layer related issues, I attempted to do the same to the Eldritch Portal, but because of how it is always rendered in full it is harder to make look good in all situations. The closest I got was using an alpha threshold:
```java
        GL11.glEnable(GL11.GL_ALPHA_TEST);
        GL11.glAlphaFunc(GL11.GL_GREATER, 0.4F);
```
to selectively render a part of the texture quad. This results in the removal of the transparent fading halo around the portal, and gives it a stricter edge, but does make it render properly layer wise.
image

I'd love to convince someone even more adept at rendering to take a look, but it's a pretty jank system. There might be a multipass rendering path, but I don't know how to do it, much less implement it via mixin.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileNodeEnergized_FixRenderingLayers.java
```diff
@@ -0,0 +1,15 @@
...
+
+import thaumcraft.api.visnet.TileVisNode;
+import thaumcraft.common.tiles.TileNodeEnergized;
+
+@Mixin(TileNodeEnergized.class)
+abstract class MixinTileNodeEnergized_FixRenderingLayers extends TileVisNode {
+
+    @Override
+    public boolean shouldRenderInPass(int pass) {
+        return pass == 1;
```
rndmorris, line 20: A comment explaining what this is intended to do would be appreciated here.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileNode_FixRenderingLayers.java
```diff
@@ -0,0 +1,16 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.common.tiles;
+
+import net.minecraft.tileentity.TileEntity;
+
+import org.spongepowered.asm.mixin.Mixin;
+
+import thaumcraft.common.tiles.TileEldritchPortal;
+
+@Mixin(TileEldritchPortal.class)
+public abstract class MixinTileNode_FixRenderingLayers extends TileEntity {
```
rndmorris:
```suggestion
public abstract class MixinTileEldritchPortal_FixRenderingLayers extends TileEntity {
```

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileNode_FixRenderingLayers.java
```diff
@@ -0,0 +1,16 @@
...
+
+    @Override
+    public boolean shouldRenderInPass(int pass) {
+        return pass == 1;
```
rndmorris, line 20: A comment here would also be beneficial.

src/main/java/dev/rndmorris/salisarcana/client/lib/RenderingQueue.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.github.bsideup.jabel.Desugar;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+
+import thaumcraft.api.aspects.AspectList;
+import thaumcraft.client.lib.RenderEventHandler;
+
+public class RenderingQueue {
+
+    public static final List<QueuedTag> tagQueue = new ArrayList<>();
```
rndmorris: All of `tagQueue`'s uses are within this class. Might be worth leaving it as package-private, if not fully private. I don't know what there would be to gain from making it public.

src/main/java/dev/rndmorris/salisarcana/client/lib/RenderingQueue.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Desugar
+    public record QueuedTag(RenderEventHandler instance, double x, double y, double z, AspectList aspects, int bright,
+        ForgeDirection dir, float partialTicks, Operation<Void> original) {}
+
+    public static void flush() {
+        if (!tagQueue.isEmpty()) {
```
rndmorris: Thought: If there are no queued tags, the iterator you're guarding here won't do much. Maybe a premature optimization?

rndmorris (changes requested):

src/main/java/dev/rndmorris/salisarcana/client/lib/RenderingQueue.java
thehoblit: removed the guard, you are right that it is redundant.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileNode_FixRenderingLayers.java
thehoblit: Sorry, this was intended to be for TileNode, not TileEldritchPortal.

thehoblit:
Made all the changes, thanks for catching the TileNode/EldritchPortal thing, I must have moved the wrong mixin from my testing project.

The same mixin with renderpass == 1 will work for  TileEldritchPortal, making it always render in front of transparent blocks instead of behind them. If you want that I will add it in pretty easily. I didn't add it to start with because the portal isn't something that is only really visible when looking directly at it, or has a good reason to be always on top like the energized node. It might be better than always being behind transparent blocks though.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileNode_FixRenderingLayers.java
```diff
@@ -0,0 +1,22 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.common.tiles;
+
+import org.spongepowered.asm.mixin.Mixin;
+
+import thaumcraft.api.TileThaumcraft;
+import thaumcraft.common.tiles.TileNode;
+
+@Mixin(TileNode.class)
+public abstract class MixinTileNode_FixRenderingLayers extends TileThaumcraft {
```
rndmorris:
```suggestion
abstract class MixinTileNode_FixRenderingLayers extends TileThaumcraft {
```

rndmorris (changes requested):
Looks fine. I'll try to test this later today.

rndmorris (approved):

PR #490, opened 2026-05-29
Add option to have unstable nodes not spawn orbs in unloaded chunks

DarkShadow44:

This PR adds a feature to prevent unstable nodes from creating aspect orbs when they are close to unloaded chunks.

Currently, unstable nodes spawn aspect orbs when they're getting ticked. Those orbs normally despawn ober time automatically.

But orbs only get updated if `World.updateEntityWithOptionalForce` passes the following check:
```
byte b0 = isForced ? (byte)0 : 32;
boolean canUpdate = !p_72866_2_ || this.checkChunksExist(i - b0, 0, j - b0, i + b0, 0, j + b0);
```

So, if orbs either
- spawn in a chunk that is not force loaded and touches unloaded chunks or
- fly into an unloaded chunk

Then the orb will never tick, and therefore never die. This leads to orbs piling up, especially since unstable nodes do randomly recharge, i.e. will never run out of aspects.

This adds a config to simply never spawn orbs if they can't be ticked - preventing infinite pileup.

No

**Other information**:

A user reported an "Oversized chunk" warning from hodgepodge, upon investigation I noticed 39666 Thaumcraft Aspect orbs. The player built a mining outpost far away from their base, chunk loaded it and the unstable node was in a chunk next to the force loaded chunks. Since Minecraft still loads neighboring chunks from time to time due to block updates, this lead to the node spawning a lot of orbs.

I tested by putting an unstable node at a chunk border, making unstable nodes spawn 10 orbs per tick and force loading only the chunk the node is in. After a few minutes there's thousands of aspect orbs in unloaded chunks.

### DIFF
--- a/docs/thaumcraft_configuration.md
+++ b/docs/thaumcraft_configuration.md
@@ -75,7 +75,7 @@ be adjusted by their modifier: 20% more often when bright, 20% less often when p
 
 ### Config Option: `unstableAspectDropsRequireLoadedChunks` (Default `true`)
 If true, unstable nodes not under the effects of a Node Stabilizer will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
-orb are loaded.
+orb are loaded. Additionally, aspect orbs will immediately despawn if they enter a chunk where they can't be ticked to further prevent their accumulation.
 
 ## Pickaxe of the Core Detection Ore Dictionary Labels (Group `tools`)
 
--- a/docs/thaumcraft_configuration.md
+++ b/docs/thaumcraft_configuration.md
@@ -74,7 +74,7 @@ If true, the rate at which tainted nodes will convert their surroundings to Tain
 be adjusted by their modifier: 20% more often when bright, 20% less often when pale, 50% less often when fading.
 
 ### Config Option: `unstableAspectDropsRequireLoadedChunks` (Default `true`)
-If true, unlocked unstable nodes will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
+If true, unstable nodes not under the effects of a Node Stabilizer will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
 orb are loaded.
 
 ## Pickaxe of the Core Detection Ore Dictionary Labels (Group `tools`)
--- a/docs/thaumcraft_configuration.md
+++ b/docs/thaumcraft_configuration.md
@@ -73,6 +73,10 @@ will scale with the current amount of vis of the node.
 If true, the rate at which tainted nodes will convert their surroundings to Tainted Lands and spawn taint tendrils will
 be adjusted by their modifier: 20% more often when bright, 20% less often when pale, 50% less often when fading.
 
+### Config Option: `unstableAspectDropsRequireLoadedChunks` (Default `true`)
+If true, unlocked unstable nodes will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
+orb are loaded.
+
 ## Pickaxe of the Core Detection Ore Dictionary Labels (Group `tools`)
 
 ### Config Option: `elementalPickOredictFilter`
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -70,8 +70,9 @@ public class ConfigThaumcraft extends ConfigGroup {
         this,
         "unstableAspectDropsRequireLoadedChunks",
         """
-            If true, unlocked unstable nodes will not drain or spawn aspect orbs unless the nearby chunks needed to
-            tick the spawned orb are loaded.""").setCategory(nodeBehaviorsCategory);
+            If true, unstable nodes not under the effects of a Node Stabilizer will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
+            orb are loaded. Additionally, aspect orbs will immediately despawn if they enter a chunk where they can't be ticked to further prevent their accumulation.""")
+            .setCategory(nodeBehaviorsCategory);
 
     //
     // Potion Ids
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
@@ -66,6 +66,13 @@ public class ConfigThaumcraft extends ConfigGroup {
         tendrils will be adjusted by their modifier: 20% more often when bright, 20% less often when pale, 50% less
         often when fading.""").setCategory(nodeBehaviorsCategory);
 
+    public final ToggleSetting unstableAspectDropsRequireLoadedChunks = new ToggleSetting(
+        this,
+        "unstableAspectDropsRequireLoadedChunks",
+        """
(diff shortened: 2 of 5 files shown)


docs/thaumcraft_configuration.md
```diff
@@ -73,6 +73,10 @@ will scale with the current amount of vis of the node.
 If true, the rate at which tainted nodes will convert their surroundings to Tainted Lands and spawn taint tendrils will
 be adjusted by their modifier: 20% more often when bright, 20% less often when pale, 50% less often when fading.
 
+### Config Option: `unstableAspectDropsRequireLoadedChunks` (Default `true`)
+If true, unlocked unstable nodes will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
```
Nikolay-Sitnikov:
```suggestion
If true, unstable nodes not under the effects of a Node Stabilizer will not drain or spawn aspect orbs unless the nearby chunks needed to tick the spawned
```

Nikolay-Sitnikov:
Does this actually cover all possible circumstances? The check for whether the orbs can spawn uses the same zone size as the "can tick" check. Therefore, there still could be circumstances where aspect orbs spawn at the node and then roll off into an adjacent block space that prevents it from ticking further. Maybe the check should have a bit of extra margin?

DarkShadow44:
> Does this actually cover all possible circumstances? The check for whether the orbs can spawn uses the same zone size as the "can tick" check. Therefore, there still could be circumstances where aspect orbs spawn at the node and then roll off into an adjacent block space that prevents it from ticking further. Maybe the check should have a bit of extra margin?

Hmm, +/- 2 chunks seemed fine for me, although it still allows extreme conditions like moving them with water or similar into unloaded chunks. Not sure what to do about that though. What do you suggest?

rndmorris:
My thoughts:
1. Preventing them from being spawned in unloaded chunks? Fine by me.
2. Before `EntityAspectOrb.onUpdate()` calls `this.moveEntity(...)` we could check if that movement would place it into an unloaded chunk and, if so, kill/despawn the orb instead?

DarkShadow44:
> My thoughts:
> 
>     1. Preventing them from being spawned in unloaded chunks? Fine by me.
> 
>     2. Before `EntityAspectOrb.onUpdate()` calls `this.moveEntity(...)` we could check if that movement would place it into an unloaded chunk and, if so, kill/despawn the orb instead?

We could add the same check used for spawning into the orb, so if it moves into a chunk where it wouldn't get updated anymore it could get despawned. Just checking for unloaded chunk won't be enough, since being in a loaded chunk doesn't guarantee being ticked.

Nikolay-Sitnikov:
> > My thoughts:
> > ```
> > 1. Preventing them from being spawned in unloaded chunks? Fine by me.
> > 
> > 2. Before `EntityAspectOrb.onUpdate()` calls `this.moveEntity(...)` we could check if that movement would place it into an unloaded chunk and, if so, kill/despawn the orb instead?
> > ```
> 
> We could add the same check used for spawning into the orb, so if it moves into a chunk where it wouldn't get updated anymore it could get despawned. Just checking for unloaded chunk won't be enough, since being in a loaded chunk doesn't guarantee being ticked.

There are two possible circumstances:
 - The orb moves into an unloaded chunk
 - The chunk unloads while an orb is within it.
 
We cannot prevent the latter situation from saving the entity (well, there is the `EntityEvent.CanUpdate` that allows us to force orbs to tick in said far-away chunks, but that doesn't help us if the chunk goes immediately from fully-loaded to completely unloaded.)

I think preventing the first situation will be good enough. Plus, even entities that don't tick can be killed, so that check could be as easy as: "after moving, check if the area around the current chunk is loaded. if not, `setDead()`"

I don't know what the performance impact of such a change would be, though. If necessary, we could optimize the check by checking if the entity crossed a chunk boundary, and if it didn't, skipping the check.

DarkShadow44:
We don't really care about a few being saved, as long as it's only a few. I added another guard to kill them if they get too close to unloaded chunks. That is just a few lookups into a hashmap, so it should be fine performance wise.

Note that just the orb mixin does not replace the spawn protection - since a node would spawn orbs into chunks where they never get updated in the first place, so the orb mixin wouldn't run

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/entities/MixinEntityAspectOrb_ChunkGuard.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import thaumcraft.common.entities.EntityAspectOrb;
+
+@Mixin(EntityAspectOrb.class)
+public abstract class MixinEntityAspectOrb_ChunkGuard extends Entity {
```
Nikolay-Sitnikov:
```suggestion
abstract class MixinEntityAspectOrb_ChunkGuard extends Entity {
```

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
```diff
@@ -66,6 +66,13 @@ public class ConfigThaumcraft extends ConfigGroup {
         tendrils will be adjusted by their modifier: 20% more often when bright, 20% less often when pale, 50% less
         often when fading.""").setCategory(nodeBehaviorsCategory);
 
+    public final ToggleSetting unstableAspectDropsRequireLoadedChunks = new ToggleSetting(
+        this,
+        "unstableAspectDropsRequireLoadedChunks",
+        """
+            If true, unlocked unstable nodes will not drain or spawn aspect orbs unless the nearby chunks needed to
+            tick the spawned orb are loaded.""").setCategory(nodeBehaviorsCategory);
```
Nikolay-Sitnikov: You should probably add more info about the new mixin to this description (and the documentation file.)

Nikolay-Sitnikov (dismissed):

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigThaumcraft.java
DarkShadow44: Updated

DarkShadow44:
Anything I should do here?

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileNode_UnstableAspectDropsChunkGuard.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
+
+import thaumcraft.api.TileThaumcraft;
+import thaumcraft.api.aspects.Aspect;
+import thaumcraft.common.tiles.TileNode;
+
+@Mixin(value = TileNode.class, remap = false)
+public abstract class MixinTileNode_UnstableAspectDropsChunkGuard extends TileThaumcraft {
```
rndmorris:
```suggestion
abstract class MixinTileNode_UnstableAspectDropsChunkGuard extends TileThaumcraft {
```

rndmorris (approved):

PR #505, opened 2026-07-22
Fix Arcane Pressure Plates rendering as full blocks

### LINKED ISSUE #504: [Bug]: Arcane Pressure Plate looks like a block
### Salis Arcana Version

2.6.0

### Where is the bug?

Salis Arcana

### Describe the bug

When using Salis Arcana 2.6.0, pressure plate render is broken in-game -> their hitbox is still good tho.
Don't know if that came with 2.6.0 or if the bug was already present in prior versions.

Image
Image

These are the config files used :

[config/salisarcana](https://github.com/user-attachments/files/29919895/salisarcana.zip)

Thanks in advance !

### To Reproduce

1. Install latest build of your mod
2. Enter whatever world you want
3. Try to place an Arcane Pressure Plate

I noticed it also does this render issue with default non-edited config files for SA

### Expected Behavior

Block should be looking like a plate, like vanilla TC4

### Attachments

_No response_

### Additional Context

_No response_

Nikolay-Sitnikov:
Bug fix - closes #504

Arcane Pressure Plates are rendered as a full block when `fixBlockBoundsAlterations` is on.

### What is the new behavior?
We duplicate the block bound logic into the renderer, making sure the block gets rendered correctly.

No

### Other Information
I justify the following `@Redirect` by pointing out that the `MixinBlockRenderer_ConserveBlockBounds` mixin which this complements also uses `@Redirect` to replace all other calls of `block.setBlockBounds`, like this mixin does.

### Checklist
- [ ] Changes have been tested using an obfuscated client connected to an obfuscated standalone server - cannot get obfuscated client to start.


rndmorris (approved):
Does what it says

PR #507, opened 2026-07-27
Create a small crater under hungry nodes when they generate

koolkrafter5:
Feature

Hungry nodes just spawn in the world and are nearly impossible to detect until it is too late and you have become node food. It takes a while for a sizeable crater to appear to warn players of the danger.

### What is the new behavior?
They break a bunch of blocks around themselves when they generate to be a bit more obvious to players without needing to wait. The number of blocks broken respects the `hungryModifierSpeed` config.
hungry_node

No


rndmorris:
I haven't had a chance to test yet. How does this interact with non-worldgen node creation? The primal focus's rare effect or the create-node command?

rndmorris (changes requested):
Hungry nodes created with the `/create-node` command do not create a crater.

Concern: hungry nodes created by using the creative-only node placer *do* create a crater. I would prefer they not, especially since it results in instantly deleting blocks around the new node. I can think of at least two modpacks that use the node placer as a quest reward.

rndmorris:
Confirmed that hungry nodes generated by the Primal wand focus also delete terrain when they spawn in.

This can be tested by temporarily adding and requiring the below mixins.
```java
@Mixin(EntityPrimalOrb.class)
abstract class MixinEntityPrimalOrb_ForceNode {

    @WrapOperation(method = "onImpact", at = @At(value = "INVOKE", target = "Ljava/util/Random;nextInt(I)I"), slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;createExplosion(Lnet/minecraft/entity/Entity;DDDFZ)Lnet/minecraft/world/Explosion;")))
    int forceSpecial(Random instance, int i, Operation<Integer> original) {
        return 0;
    }

    @WrapOperation(method = "onImpact", at =@At(value = "INVOKE", target = "Ljava/util/Random;nextBoolean()Z"), slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;createExplosion(Lnet/minecraft/entity/Entity;DDDFZ)Lnet/minecraft/world/Explosion;")))
    boolean forceNode(Random instance, Operation<Boolean> original) {
        return false;
    }
}

@Mixin(value = ThaumcraftWorldGenerator.class, remap = false)
abstract class MixinThaumcraftWorldGenerator_ForceHungryNode {
    @WrapOperation(method = "createRandomNodeAt", at = @At(value = "INVOKE", target = "Ljava/util/Random;nextInt(I)I", ordinal = 0))
    private static int forceSpecial(Random instance, int i, Operation<Integer> original) {
        return 0;
    }
    @WrapOperation(method = "createRandomNodeAt", at = @At(value = "INVOKE", target = "Ljava/util/Random;nextInt(I)I", ordinal = 1))
    private static int forceHungry(Random instance, int i, Operation<Integer> original) {
        return 9;
    }
}
```

koolkrafter5:
Sorry for the delay. Should be fixed now. Tested with the node placer. It did break the blocks before and now it does not.

PR #508, opened 2026-08-04
Use bindTexture cache

boubou19:
Improve memory allocations

Currently, `UtilsFX.bindTexture` is checking if the string is inside the map, but if it's not, it does not populate the cache, so the cache is never used.

### What is the new behavior?
Make it so the cache is populated and work as intended.

No.

### Checklist

### DIFF
--- a/docs/bugfixes.md
+++ b/docs/bugfixes.md
@@ -388,6 +388,7 @@ Lowers the frequency of Thaumcraft's Arcane Bore calls to drain vis for speedup,
 **Config option:** `fixBindTextureCache`
 
 Eliminates the per-frame ResourceLocation allocations by making UtilsFX populate its bound texture cache.
+
 ## Allow Sapling Drops from Connected Magical Leaves
 
 **Config option:** `allowDropsFromLiveLeaves`
--- a/docs/bugfixes.md
+++ b/docs/bugfixes.md
@@ -376,3 +376,9 @@ Fixes a bug where some GUIs would render the wrong aspects when shifting over a 
 **Config option:** `boreDecreaseCVisCheckFrequency`
 
 Lowers the frequency of Thaumcraft's Arcane Bore calls to drain vis for speedup, and therefore calls to find vis nets.
+
+## Fix bound texture cache not being populated
+
+**Config option:** `fixBindTextureCache`
+
+Eliminates the per-frame ResourceLocation allocations by making UtilsFX populate its bound texture cache.
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinUtilsFX_BindTextureCache.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinUtilsFX_BindTextureCache.java
@@ -1,0 +1,36 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.client.lib;
+
+import java.util.Map;
+
+import net.minecraft.util.ResourceLocation;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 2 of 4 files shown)


Nikolay-Sitnikov:
Boubou, small question: is this actually a performance benefit? From what I've heard, Java can optimize out non-escaping allocations pretty well. I'm not sure if it can accomplish it in this case (I could try logging Java native code output to check), but it if can, then this caching step would actually *worsen* performance.

It might be better to just replace the `.containsKey` check with `false`. (Especially considering that the "cached" path doesn't actually make use of `computeIfAbsent`, meaning that 2 hash-map lookups have to occur instead of one allocation. Plus, `boundTextures` isn't actually thread-safe, so this could theoretically lead to transient NPEs if one thread decides to resize the hash map while another is looking up an entry.)

boubou19:
> Boubou, small question: is this actually a performance benefit? From what I've heard, Java can optimize out non-escaping allocations pretty well. I'm not sure if it can accomplish it in this case (I could try logging Java native code output to check), but it if can, then this caching step would actually _worsen_ performance.

If i did the PR, it's because i saw it on the profile and it's a quick fix. Now you can check for yourself, but this made the thing disappear so for me it's good enough.

> It might be better to just replace the `.containsKey` check with `false`. (Especially considering that the "cached" path doesn't actually make use of `computeIfAbsent`, meaning that 2 hash-map lookups have to occur instead of one allocation. Plus, `boundTextures` isn't actually thread-safe, so this could theoretically lead to transient NPEs if one thread decides to resize the hash map while another is looking up an entry.)

Well as i said, my issue disappeared, so i do not want to enter the microoptimization domain. About the threading issues, i tested it in full NH pack and i did not get such issues. Now CME / random NPEs like these will be easy to fix if they arise, but for now they don't, and this fixes what seems to be an oversight from the original code intent.

docs/bugfixes.md
```diff
@@ -383,6 +383,11 @@ Fixes a bug where some GUIs would render the wrong aspects when shifting over a
 
 Lowers the frequency of Thaumcraft's Arcane Bore calls to drain vis for speedup, and therefore calls to find vis nets.
 
+## Fix bound texture cache not being populated
+
+**Config option:** `fixBindTextureCache`
+
+Eliminates the per-frame ResourceLocation allocations by making UtilsFX populate its bound texture cache.
 ## Allow Sapling Drops from Connected Magical Leaves
```
Nikolay-Sitnikov, line 392:
```suggestion

## Allow Sapling Drops from Connected Magical Leaves
```

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/client/lib/MixinUtilsFX_BindTextureCache.java
```diff
@@ -0,0 +1,36 @@
...
+abstract class MixinUtilsFX_BindTextureCache {
+
+    @Shadow(remap = false)
+    private static Map<String, ResourceLocation> boundTextures;
+
+    @Inject(method = "bindTexture(Ljava/lang/String;)V", at = @At("HEAD"))
+    private static void salisarcana$cacheTexture(String texture, CallbackInfo ci) {
+        if (!boundTextures.containsKey(texture)) {
+            boundTextures.put(texture, new ResourceLocation("thaumcraft", texture));
+        }
```
Nikolay-Sitnikov: Have you considered using `computeIfAbsent`? You should be able to do it without any allocations as long as you don't capture any variables from the outside method.

```suggestion
        boundTextures.computeIfAbsent(texture, key -> new ResourceLocation("thaumcraft", key));
```

Nikolay-Sitnikov (changes requested):

Nikolay-Sitnikov:
Also, considering this is a performance-critical method and the method is really simple, you could probably justify using `@Overwrite` here.

Nikolay-Sitnikov (approved):
There could still be improvements made, but this is acceptable.

PR #512, opened 2026-08-07
Quick-stash Wand Foci

### LINKED ISSUE #465: [Enhancement]: Quickly add foci to focus pouches
### Salis Arcana Version

v2.5.0

### Context

Making a new focus means you need to either take your focus pouch off to add it or remove the focus from a wand in your inventory to add it to your focus pouch.

### Requested Feature

Press the Wand Focus keybind (default F) while holding a wand focus to quickly put it into your focus pouch.

### Alternative Solutions

Add a feature that comes up with alternative solutions for this issue template automatically.

### Additional Context

_No response_

Nikolay-Sitnikov:
Feature - adds a new keybind to stash a Wand Focus out of an open inventory or your hand into your Focus Pouch(es).

Such a keybind does not exist. Closes #465.

### What is the new behavior?
Pressing "F" (configurable like any other hotkey) will take a focus out of any slot, if that slot is accessible to the player, and store it in their equipped Focus Pouches, using the same order as vanilla Thaumcraft.

When pressed in-game, instead stores the wand focus held in the player's hand.

No. However, GTNH likes making it so all modded keybinds are unset by default (ex. [MixinKeyHandlerThaumcraft](https://github.com/GTNewHorizons/Hodgepodge/blob/master/src/main/java/com/mitchej123/hodgepodge/mixins/late/thaumcraft/MixinKeyHandlerThaumcraft.java)), so we may need to introduce a config setting to make the keybind default to unset.

### Other information
- All inventory changes run on the client before being sent to the server. This means that stashing foci is always instant, no matter how much ping you have to the server.
- Unfortunately, implementing a GUI key handler requires a mixin in vanilla Minecraft. `MixinGuiContainer_HandleKeyTyped` does the job when NEI is not installed, and we use `IContainerInputHandler` when NEI is installed.
- The TODO in `QuickStashFocus` was left there intentionally. If you believe it to be an important feature, I can implement it relatively simply.
- I intentionally did not gate the entries in `Mixins.java` on the config, since we probably want to eventually allow the player (and any servers they join) to turn this feature on & off without restarting the game.

### Checklist
- [ ] New entries in `Mixins.java` are linked to a config entry so they can be enabled/disabled
- [ ] Changes have been tested using an obfuscated client connected to an obfuscated standalone server

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
@@ -15,7 +15,7 @@ public enum Mixins implements IMixins {
     // spotless:off
     // Early Mixins
     ACCESSORS(new SalisBuilder(Phase.EARLY)
-        .addClientMixins("accessor.AccessorGuiContainer", "accessor.AccessorMinecraft")),
+        .addClientMixins("accessor.AccessorGuiContainer", "accessor.AccessorMinecraft", "accessor.AccessorCreativeSlot")),
     VANILLA_GUI_KEY_TYPED(new SalisBuilder(Phase.EARLY)
         .addClientMixins("gui.MixinGuiContainer_HandleKeyTyped")
         .addExcludedMod(TargetedMod.NOT_ENOUGH_ITEMS)),
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
@@ -15,8 +15,15 @@ public enum Mixins implements IMixins {
     // spotless:off
     // Early Mixins
     ACCESSORS(new SalisBuilder(Phase.EARLY)
-        .addCommonMixins("accessor.AccessorGuiContainer")
-        .addClientMixins("accessor.AccessorMinecraft")),
+        .addClientMixins("accessor.AccessorGuiContainer", "accessor.AccessorMinecraft")),
+    VANILLA_GUI_KEY_TYPED(new SalisBuilder(Phase.EARLY)
+        .addClientMixins("gui.MixinGuiContainer_HandleKeyTyped")
+        .addExcludedMod(TargetedMod.NOT_ENOUGH_ITEMS)),
+
+    // Late Accessors
+    THAUM_ACCESSORS(new SalisBuilder()
+        .addCommonMixins("thaumcraft.common.container.AccessorContainerFocusPouch")
+        .addRequiredMod(TargetedMod.THAUMCRAFT)),
 
     // Bugfixes
     ADVANCED_ARCANE_FURNACE_SAVE_NBT(new SalisBuilder()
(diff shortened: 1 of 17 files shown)


Nikolay-Sitnikov:
Works on singleplayer & dedicated server, with or without NEI. Could not test with obfuscated client & server, since that still doesn't work for me.

Nikolay-Sitnikov:
To run this PR without NEI, mark all dependencies except Thaumcraft & Baubles Expanded as `compileOnly`, then add this:
```groovy
dependencies {
    configurations.runtimeClasspath {
        exclude(group: "com.github.GTNewHorizons", module: "NotEnoughItems")
    }
}
```
to prevent Baubles Expanded from pulling in NEI transitively. (I have also submitted a PR to remove NEI from the Gradle runtime dependencies of Baubles Expanded in the future.)

rndmorris (changes requested):
The stashing ability works great.

**Issue:** Having the stash bound to the same key as Thaumcraft's "Change Wand Focus" prevents you from changing wand focuses (the selection radial doesn't open). Disabling the feature or assigning them to different keys works around this conflict.

Nikolay-Sitnikov:
This is a vanilla bug - when you set two keybinds to the same key, only one can even get activated. There is no way around this (see `KeyBinding#setKeyBindState` for details - `KeyBinding` uses a hash map, keyed on the key code, to choose which keybind to activate. Therefore, only one keybind per key code will be activated.)

Both [Hodgepodge](https://github.com/GTNewHorizons/Hodgepodge/blob/master/src/main/java/com/mitchej123/hodgepodge/mixins/early/minecraft/MixinKeyBinding.java) & [Not Enough Items](https://github.com/GTNewHorizons/NotEnoughItems/blob/master/src/main/java/codechicken/nei/asm/KeyBindingFixHooks.java) include fixes for this issue, so I'm not sure what the right action to take here is. I guess we set the keybind default to NONE when neither of these mods are loaded? Or we could just piggy-back the Thaumcraft keybind?

rndmorris:
> Or we could just piggy-back the Thaumcraft keybind?

I think this makes a degree of sense, re-using the focus-related keybind for more focus-related activities. Can you think of any compelling reasons we shouldn't? I'm not having much luck thinking of any.

Nikolay-Sitnikov:
Well, mainly just so people could use different bindings for these, especially if they use a mouse button for switching foci on a wand.

I think I will reuse the Thaumcraft keybind only if those other mods are not loaded.

src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
```diff
@@ -15,8 +15,19 @@ public enum Mixins implements IMixins {
...
     ACCESSORS(new SalisBuilder(Phase.EARLY)
-        .addCommonMixins("accessor.AccessorGuiContainer")
-        .addClientMixins("accessor.AccessorMinecraft")),
+        .addClientMixins("accessor.AccessorGuiContainer", "accessor.AccessorMinecraft", "accessor.AccessorCreativeSlot")),
+    VANILLA_GUI_KEY_TYPED(new SalisBuilder(Phase.EARLY)
+        .addClientMixins("gui.MixinGuiContainer_HandleKeyTyped")
+        .addExcludedMod(TargetedMod.NOT_ENOUGH_ITEMS)),
+
+    // Late Accessors
+    THAUM_ACCESSORS(new SalisBuilder()
```
rndmorris, line 27: Documentation: A brief comment here to explain why we're making these mixins always-on would probably be prudent.

rndmorris (changes requested):
Otherwise it looks and behaves fine

rndmorris (approved):

PR #514, opened 2026-08-13
Preserve tile entity data during Thaumcraft harvesting

DreamYao520:
Thaumcraft removes blocks before calling their harvest logic. Blocks whose drops depend on a tile entity can therefore lose their metadata and NBT. For example, mining a Vibrant Capacitor Bank with the Primal Crusher's area mining converted it into a Chaotic Capacitor Bank.

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
+++ b/src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
@@ -130,6 +130,11 @@ public class ConfigBugfixes extends ConfigGroup {
         "equalTradeBreaksBlocks",
         "Fixes a bug where you couldn't break blocks if you were holding the equal trade focus item.");
 
+    public final ToggleSetting preserveTileEntityDrops = new ToggleSetting(
+        this,
+        "preserveTileEntityDrops",
+        "Preserves tile entity data while Thaumcraft harvests blocks, preventing the Primal Crusher from creating incorrect drops.");
+
     public final ToggleSetting nodesRechargeInGameTime = new ToggleSetting(
         this,
         "nodesRechargeInGameTime",
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/utils/MixinBlockUtils_PreserveTileEntityDrops.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/utils/MixinBlockUtils_PreserveTileEntityDrops.java
@@ -1,0 +1,41 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.common.lib.utils;
+
+import net.minecraft.block.Block;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.world.World;
+
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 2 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
```diff
@@ -130,6 +130,11 @@ public class ConfigBugfixes extends ConfigGroup {
         "equalTradeBreaksBlocks",
         "Fixes a bug where you couldn't break blocks if you were holding the equal trade focus item.");
 
+    public final ToggleSetting preserveTileEntityDrops = new ToggleSetting(
+        this,
+        "preserveTileEntityDrops",
+        "Preserves tile entity data while Thaumcraft harvests blocks, preventing the Primal Crusher from creating incorrect drops.");
```
Nikolay-Sitnikov, line 136: This config also needs to be added to `docs/bugfixes.md`. Feel free to just copy-paste the description.

Nikolay-Sitnikov (changes requested):

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/utils/MixinBlockUtils_PreserveTileEntityDrops.java
```diff
@@ -0,0 +1,41 @@
...
+
+        Block block = world.getBlock(x, y, z);
+        int metadata = world.getBlockMetadata(x, y, z);
+        block.onBlockHarvested(world, x, y, z, metadata, player);
+
+        boolean removed = block.removedByPlayer(world, player, x, y, z, true);
+        if (removed) {
+            block.onBlockDestroyedByPlayer(world, x, y, z, metadata);
+        }
+        return removed;
```
Nikolay-Sitnikov, line 42: This mixin is tricky enough that I think a little bit of code elaboration would be necessary.

Firstly, this is basically an overwrite. You should document what exactly is different between this method and `removeBlock` - specifically, that you are calling `removedByPlayer` with `true` as the `canHarvest` parameter, unlike Thaumcraft's `removeBlock`.

Secondly, you should probably specify that we know that `block` is not null, since if it was, then `canHarvest` would be false.

As far as I can tell, this mixin is correct, though. Good work.

(Theoretically, you could also optimize the code by borrowing the `block` and `metadata`/`i1` locals from the `harvestBlock` method instead of re-getting them from the world, but that's a minor optimization that might impact readability.)

src/main/java/dev/rndmorris/salisarcana/config/group/ConfigBugfixes.java
DreamYao520: added

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/lib/utils/MixinBlockUtils_PreserveTileEntityDrops.java
DreamYao520: added comment on line 30

Nikolay-Sitnikov (approved):
Works as intended.

PR #517, opened 2026-08-14
Force reservoirs to use the correct face when checking their essentia type

### LINKED ISSUE #516: [Bug]: Reservoirs don't work with directly connected unlabeled jars
### Salis Arcana Version

2.4.0

### Where is the bug?

Thaumcraft 4

### Describe the bug

Here's 4 reservoirs, each pointed down, and each containing 256 terra.
Image
I've connected warded jars to each of them, in different ways. Jars have higher suction than reservoirs, so they should all drain 64 esssentia from their reservoir...
But the one that's unlabeled and connected directly to the reservoir didn't.
Image
It turns out that the code for taking essentia out of a connected block has to specify which aspect it's taking. So in the case of generic suction, it has to ask what aspect is in that other block and then take 1 essentia of that aspect out of the other block. Reservoirs are coded weirdly and don't say what aspect(s) they have unless you ask in a specific way, and the only block in vanilla thaum that asks that way is the essentia tube.

### To Reproduce

See above.

### Expected Behavior

All 4 jars should drain essentia from their reservoir.

### Attachments

_No response_

### Additional Context

`TileEssentiaReservoir.getEssentiaType()` might've been intended to check whether `loc == face` instead of `loc == ForgeDirection.UNKNOWN`?

rndmorris:
Bugfix

Closes #516 

### What is the new behavior?
Reservoirs will check the input direction instead of `UNKNOWN`.

No

### Other information
"Run Obfuscated Client" loads no mods, not even Salis Arcana, on my machine.

### Checklist
- [ ] Changes have been tested using an obfuscated client connected to an obfuscated standalone server

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
@@ -325,8 +325,7 @@ public enum Mixins implements IMixins {
     RESERVOIR_GETESSENTIATYPE_USES_ARG(new SalisBuilder()
         .applyIf(SalisConfig.bugfixes.reservoirsUseArgInGetEssentiaType)
         .addCommonMixins("thaumcraft.common.tiles.MixinTileEssentiaReservoir_UseArgInGetEssentiaType")
-        .addRequiredMod(TargetedMod.THAUMCRAFT)
-    ),
+        .addRequiredMod(TargetedMod.THAUMCRAFT)),
 
     // Features
     EXTENDED_BAUBLES_SUPPORT(new SalisBuilder()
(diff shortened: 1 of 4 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/Mixins.java
```diff
@@ -315,6 +315,11 @@ public enum Mixins implements IMixins {
         .addCommonMixins("thaumcraft.common.blocks.MixinBlockMagicalLeaves_AllowConnectedLeafDrops")
         .addRequiredMod(TargetedMod.THAUMCRAFT)),
 
+    RESERVOIR_GETESSENTIATYPE_USES_ARG(new SalisBuilder()
+        .applyIf(SalisConfig.bugfixes.reservoirsUseArgInGetEssentiaType)
+        .addCommonMixins("thaumcraft.common.tiles.MixinTileEssentiaReservoir_UseArgInGetEssentiaType")
+        .addRequiredMod(TargetedMod.THAUMCRAFT)
+    ),
```
Nikolay-Sitnikov:
```suggestion
        .addRequiredMod(TargetedMod.THAUMCRAFT)),
```

Nikolay-Sitnikov (changes requested):
Fix this, fix the merge conflict, and this is ready to approve.

Nikolay-Sitnikov (approved):

PR #520, opened 2026-08-18
Speed up potion aspect registration by only computing the effect lists once

boubou19:
This PR fixes a performance issue during boot.

Thaumcraft computes the effect list of each potion, computing the effect list for 32767 potions

### What is the new behavior?
Potions have bit formulas, and it is possible to compute a mask to discard all the bits that are not considered for any potion.
From there, we can compute potion effect lists once then cache them.

This is what this PR does. It also compute the bitmask dynamically in case some mod changes the potion effect list or something, and fallback to original behavior if no meaningful bitmask can be computed.

In comparison, in latest dev builds of NH, Thaumcraft postinit goes down from 1.1s to only 0.3s with this change.

No

### Checklist

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
+++ b/src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
@@ -6,7 +6,7 @@ import java.util.function.IntFunction;
 
 public final class PotionMetadataCache<T> {
 
-    public static final int ALL_METADATA_BITS = 0x7FFF;
+    public static final int ALL_METADATA_BITS = 32767; // 2^15-1
     private static final int SPLASH_BIT = 1 << 14;
 
     private final int metadataMask;
@@ -17,6 +17,7 @@ public final class PotionMetadataCache<T> {
     }
 
     public T get(int metadata, IntFunction<T> loader) {
+        // Metadata values that differ only in irrelevant bits produce identical effects.
         int key = metadata & metadataMask;
         if (!values.containsKey(key)) {
             values.put(key, loader.apply(metadata));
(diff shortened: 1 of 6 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/config/MixinConfigAspects_SpeedupPotionAspects.java
```diff
@@ -0,0 +1,52 @@
...
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
+
+import dev.rndmorris.salisarcana.lib.PotionMetadataCache;
+import thaumcraft.common.config.ConfigAspects;
+
+@Mixin(value = ConfigAspects.class, remap = false)
+abstract class MixinConfigAspects_SpeedupPotionAspects {
+
+    @Unique
+    private static PotionMetadataCache<List<PotionEffect>> salisarcana$potionEffectCache;
```
Nikolay-Sitnikov: You could use an `@Share` to store this in a local variable within the method instead of making an entirely separate static field for it.

Nikolay-Sitnikov:
Also, boubou, is there any documentation for the format of `potionRequirements` and what it means?

Nikolay-Sitnikov:
Also, it might be easier to accomplish your goal by generating the aspects on-the-fly within `ThaumcraftRecipeManager` and `ThaumcraftApi` rather than trying to optimize that loop. After all, most people won't need to get the aspects of every single potion bottle in their playthroughs.

boubou19:
> is there any documentation for the format of `potionRequirements` and what it means?

Not really, you need to go through MC code to find infos about it, like in `PotionHelper.parsePotionEffects` or in the definition of the effect strings in the begining of the class, for instance, `public static final String ghastTearEffect = "+0-1-2-3&4-4+13";`

> Also, it might be easier to accomplish your goal by generating the aspects on-the-fly within `ThaumcraftRecipeManager` and `ThaumcraftApi` rather than trying to optimize that loop. After all, most people won't need to get the aspects of every single potion bottle in their playthroughs.

I’m not sure about that approach. Lazy aspect generation would change Thaumcraft’s registration behavior and touch multiple lookup paths, which seems riskier than preserving vanilla behavior and only optimizing the expensive loop. My goal with this PR is to speed up the incredible time taken on just potion iterations. Saving 0.8s of boot time on NH when my game boots already faster than a minute is quite the saving.

src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public T get(int metadata, IntFunction<T> loader) {
+        int key = metadata & metadataMask;
+        if (!values.containsKey(key)) {
+            values.put(key, loader.apply(metadata));
+        }
+        return values.get(key);
+    }
+
+    public static int findRelevantBits(Iterable<?> requirements, Iterable<?> amplifiers) {
```
Nikolay-Sitnikov: `requirements` and `amplifiers` are all maps of `Integer` to `String`, and the vanilla code assumes that. Make these `Iterable<String>`, toss a `@SuppressWarnings("unchecked")` on the mixin, and skip the type-checking code in `addExpressionBits`.

src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
```diff
@@ -0,0 +1,65 @@
...
+    public PotionMetadataCache(int metadataMask) {
+        this.metadataMask = metadataMask;
+    }
+
+    public T get(int metadata, IntFunction<T> loader) {
+        int key = metadata & metadataMask;
+        if (!values.containsKey(key)) {
+            values.put(key, loader.apply(metadata));
+        }
+        return values.get(key);
```
Nikolay-Sitnikov, line 25: Why not use `computeIfAbsent`?

src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public static int findRelevantBits(Iterable<?> requirements, Iterable<?> amplifiers) {
+        int result = addExpressionBits(SPLASH_BIT, requirements);
+        return result == ALL_METADATA_BITS ? result : addExpressionBits(result, amplifiers);
+    }
+
+    private static int addExpressionBits(int bits, Iterable<?> expressions) {
```
Nikolay-Sitnikov: This function could really use a plethora of comments explaining what information we are extracting from `potionRequirements` & `potionAmplifiers`.

Nikolay-Sitnikov (changes requested):

src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
Nikolay-Sitnikov: Also, if the intent is that `metadata` and `metadata & metadataMask` (which is `key`) are semantically equivalent, you should be able to just do `k -> loader.apply(k)` in `computeIfAbsent`, without needing to capture `metadata` from the outer scope.

(And if you change `loader` to be a `Function<Integer, T>`, then you could skip the per-get allocations altogether and just pass `loader` into `computeIfAbsent`.)

boubou19: You are right about the key caching being more explicit. But about the `computeIfAbsent`, it cannot be used directly here, as `PotionHelper.getPotionEffects` returns null as valid output which means "no effects". So you'd need to wrap it with a `Optional.ofNullable` just to use the `computeIfAbsent` variant. I find the current version cleaner.

src/main/java/dev/rndmorris/salisarcana/lib/PotionMetadataCache.java
boubou19: should be much clearer now.

Nikolay-Sitnikov (approved):
Looks good, as far as I can tell.

PR #524, opened 2026-08-25
Make Wand Focus: Warding treat the world metadata as authoritative

Nikolay-Sitnikov:
Bug fix

Closes GTNewHorizons/Dupes-Exploits-GTNH#348

Using Wand Focus: Warding on a block with a metadata value > 127 would cause the metadata to get truncated to a byte when the block was un-warded, allowing for blocks to be transmuted between metadata.

### What is the new behavior?
Wand Focus: Warding already sets the metadata of the Warded Block to the un-truncated metadata of the original block in order for the block renderer to work, so this PR changes the un-warding part of Wand Focus: Warding to read the metadata of the world rather than the one in the NBT.

Since this also means that the metadata in the NBT is now useless, introduces a second option to entirely remove that metadata (at the cost of making those save files *require* the first mixin to be enabled.)

No, unless you enable that second mixin.

### Other information
There seems to be an unrelated bug with EndlessIDs that causes Warded Blocks to render as transparent when you first ward a block (and render normally after you trigger the chunk to re-render a second time), unrelated to these mixins.

There also seems to be yet another unrelated bug in that GT5U ores don't actually render the "ore" part of their textures when warded, but that also seems to be unrelated to these mixins.

### Checklist
- [x] Config entries are in the appropriate group (if unclear where something belongs, ask) - I chose to put both settings in the same group, even though the second setting is more of an optimization, so they'd be located next to each other in the config file.
- [ ] Changes have been tested using an obfuscated client connected to an obfuscated standalone server - not on an obfuscated client, but with GT5U

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/items/wands/foci/MixinItemFocusWarding_UseWorldMetadata.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/items/wands/foci/MixinItemFocusWarding_UseWorldMetadata.java
@@ -1,0 +1,29 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.common.items.wands.foci;
+
+import net.minecraft.world.World;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
+
+import com.llamalad7.mixinextras.sugar.Local;
+
+import thaumcraft.api.BlockCoordinates;
+import thaumcraft.common.items.wands.foci.ItemFocusWarding;
+
+@Mixin(ItemFocusWarding.class)
+abstract class MixinItemFocusWarding_UseWorldMetadata {
+
+    @ModifyArg(
+        method = "onFocusRightClick",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/World;setBlock(IIILnet/minecraft/block/Block;II)Z",
+            ordinal = 1,
+            remap = true),
+        index = 4,
+        remap = false)
+    private int getMetadataFromWorld(int metadata, @Local(argsOnly = true) World world, @Local BlockCoordinates c) {
+        return world.getBlockMetadata(c.x, c.y, c.z);
+    }
+}
--- a/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileWarded_DontStoreMetadata.java
+++ b/src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileWarded_DontStoreMetadata.java
@@ -1,0 +1,34 @@
+package dev.rndmorris.salisarcana.mixins.late.thaumcraft.common.tiles;
+
+import net.minecraft.nbt.NBTTagCompound;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import com.llamalad7.mixinextras.expression.Definition;
+import com.llamalad7.mixinextras.expression.Expression;
+
+import thaumcraft.api.TileThaumcraft;
+import thaumcraft.common.tiles.TileWarded;
+
+@Mixin(TileWarded.class)
+abstract class MixinTileWarded_DontStoreMetadata extends TileThaumcraft {
+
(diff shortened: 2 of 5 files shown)


src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/items/wands/foci/MixinItemFocusWarding_UseWorldMetadata.java
```diff
@@ -0,0 +1,29 @@
...
+    @ModifyArg(
+        method = "onFocusRightClick",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/World;setBlock(IIILnet/minecraft/block/Block;II)Z",
+            ordinal = 1,
+            remap = true),
+        index = 4,
+        remap = false)
+    private int getMetadataFromWorld(int metadata, @Local(argsOnly = true) World world, @Local BlockCoordinates c) {
```
rndmorris, line 26: Nitpick: I'm seeing a `@Local can use variable name` compiler warning on `@Local BlockCoordinates c`.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileWarded_DontStoreMetadata.java
```diff
@@ -0,0 +1,34 @@
...
+import com.llamalad7.mixinextras.expression.Definition;
+import com.llamalad7.mixinextras.expression.Expression;
+
+import thaumcraft.api.TileThaumcraft;
+import thaumcraft.common.tiles.TileWarded;
+
+@Mixin(TileWarded.class)
+abstract class MixinTileWarded_DontStoreMetadata extends TileThaumcraft {
+
+    @Definition(id = "getByte", method = "Lnet/minecraft/nbt/NBTTagCompound;getByte(Ljava/lang/String;)B", remap = true)
```
rndmorris, line 18: Nitpick: `remap = true` looks to be redundant here.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileWarded_DontStoreMetadata.java
```diff
@@ -0,0 +1,34 @@
...
+    @Expression("?.getByte('md')")
+    @Redirect(method = "readCustomNBT", at = @At("MIXINEXTRAS:EXPRESSION"), remap = false)
+    private byte getMetadataFromWorld(NBTTagCompound instance, String s) {
+        return (byte) this.worldObj.getBlockMetadata(this.xCoord, this.yCoord, this.zCoord);
+    }
+
+    @Definition(
+        id = "setByte",
+        method = "Lnet/minecraft/nbt/NBTTagCompound;setByte(Ljava/lang/String;B)V",
+        remap = true)
```
rndmorris, line 28: Nitpick: and also here.

rndmorris (changes requested):
Very small nitpicks, otherwise looks good.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/items/wands/foci/MixinItemFocusWarding_UseWorldMetadata.java
Nikolay-Sitnikov: I don't agree with that lint. If there's only one variable of the selected type, targeting it by type is more dependable than targeting it by its name (which wouldn't even work if it was vanilla, since variable names don't exist in obfuscated vanilla classes.) Also, it would just look messier for no benefit.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileWarded_DontStoreMetadata.java
Nikolay-Sitnikov: Just checked: without this `remap = true`, the mixin fails to apply & crashes when run on an obfuscated client.

src/main/java/dev/rndmorris/salisarcana/mixins/late/thaumcraft/common/tiles/MixinTileWarded_DontStoreMetadata.java
Nikolay-Sitnikov: Just checked: without this `remap = true`, the mixin fails to apply & crashes when run on an obfuscated client.

rndmorris (approved):

PR #529, opened 2026-08-31
Fix quick-stash keybind resetting on every load

### LINKED ISSUE #527: [Bug]: Quick-Stash Focus keybind resets to F on every game start
### Salis Arcana Version

1.1.70-GTNH (also reproduced on 1.1.68-GTNH). GTNH daily   708, Minecraft 1.7.10.

### Where is the bug?

Salis Arcana

### Describe the bug

The Quick-Stash Focus keybind is saved to options.txt correctly but is never loaded back, so it reverts to its default F on every launch.

The cause is the KeyBinding's description string, which contains a colon: salisarcana:keybind.quick_stash_focus (seen in dev/rndmorris/salisarcana/client/QuickStashFocus).

options.txt is a colon-delimited format, and the loader compares only the part before the first colon. In GTNH, GameSettings.loadOptions is replaced by a mixin from BugTorch (jss/bugtorch/mixins/early/minecraft/optimization/gamesettings/MixinFasterLoadOptions), which does this:

    astring = line.split(":");
    if (astring[0].startsWith("key_")) {
        for (KeyBinding kb : this.keyBindings) {
            if (astring[0].equals("key_" + kb.getKeyDescription()))
                kb.setKeyCode(Integer.parseInt(astring[1]));
        }
    }

For this binding the saved line is

    key_salisarcana:keybind.quick_stash_focus:72

so astring[0] is "key_salisarcana", which never equals "key_salisarcana:keybind.quick_stash_focus". The stored keycode is therefore never applied and the KeyBinding keeps the default it was registered with (KEY_F). Vanilla's own loadOptions has the same shape, so this is not specific to BugTorch.

saveOptions writes the full description, so the file always looks correct. That asymmetry is what makes the bug confusing to diagnose: the file says one key, the game uses another.

Nothing appears in the log, because no exception is thrown. Integer.parseInt is only reached on a match, so there is no "Skipping bad option" warning either.

### To Reproduce

1. Leave quickStashFocus enabled in config/salisarcana/enhancements.cfg (default true).
2. In Options -> Controls, rebind Quick-Stash Focus from F to some other key, e.g. Numpad 8.
3. Check options.txt. It now correctly reads key_salisarcana:keybind.quick_stash_focus:72 (72 is LWJGL KEY_NUMPAD8).
4. Restart the game.
5. The keybind is back on F, while options.txt still holds 72.

### Expected Behavior

The chosen key persists across restarts, like every other keybind.

### Attachments

The relevant line from options.txt:

    key_salisarcana:keybind.quick_stash_focus:72

It is the only line in the file whose key name contains a colon. For comparison, BogoSorter's line

    key_key.bogosorter.pin_slot:-98:CONTROL

also has three fields, but its colon is on the value side (a modifier), so astring[0] still matches and astring[1] still parses. Only a colon inside the key description breaks the lookup.

### Additional Context

Suggested fix: give the KeyBinding a description without a colon, e.g. salisarcana.keybind.quick_stash_focus, and rename the matching lang key.

Worth a changelog note: after such a rename the existing options.txt entries become orphaned, so everyone's binding will reset to the default once.

Practical impact in GTNH: F is also NEI's "focus search field" key (key_nei.options.keys.gui.search), so this binding silently takes it back on every launch and the NEI search box stops responding to F until the player notices and rebinds it again.

Nikolay-Sitnikov:
Bug fix

Closes #527

### What is the new behavior?
Removed colons from the lang keys for the keybinds, so Minecraft parses them correctly.

No. It would have reset everyone's keybinds for quick-stashing foci, but that keybind already resets itself on every load.

### Checklist

### DIFF
--- a/src/main/java/dev/rndmorris/salisarcana/client/QuickStashFocus.java
+++ b/src/main/java/dev/rndmorris/salisarcana/client/QuickStashFocus.java
@@ -27,9 +27,9 @@ public final class QuickStashFocus {
         // If none of these mods are present, use the Thaumcraft keybind instead.
         if (MixinModCompat.multiKeyBindsPermitted()) {
             KEYBIND = new KeyBinding(
-                "salisarcana:keybind.quick_stash_focus",
+                "salisarcana.keybind.quick_stash_focus",
                 Keyboard.KEY_F,
-                "salisarcana:keybind_category");
+                "salisarcana.keybind_category");
             ClientRegistry.registerKeyBinding(KEYBIND);
         }
 
--- a/src/main/resources/assets/salisarcana/lang/en_US.lang
+++ b/src/main/resources/assets/salisarcana/lang/en_US.lang
@@ -174,8 +174,9 @@ salisarcana:duplicate_research.text=Duplicate "%s"
 salisarcana:duplicate_research.prompt=Shift-click to purchase a duplicate of this research
 salisarcana:duplicate_research.success=Duplicated research paper added to inventory
 
-salisarcana:keybind_category=Salis Arcana
-salisarcana:keybind.quick_stash_focus=Quick-Stash Focus
+# Keybinds - these lang keys cannot contain colons, or Minecraft won't parse them correctly when loading.
+salisarcana.keybind_category=Salis Arcana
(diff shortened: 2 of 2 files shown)


rndmorris (approved):
Does what it says

PR #530, opened 2026-09-12
Make Wand Foci emit block break & block place events

Nikolay-Sitnikov:
Bug fix / security improvement.

- Wand Focus: Fire upgraded with Alchemist's Fire can place Fire blocks in claimed chunks
- Wand Focus: Shock upgraded with Earth Shock can place lightning blocks in claimed chunks
- Wand Focus: Equal Trade can swap blocks within claimed chunks (both directly and via flood fill)
- Wand Focus: Warding can ward & unward blocks within claimed chunks.

(initially reported in GTNewHorizons/Dupes-Exploits-GTNH#364.)

### What is the new behavior?
Added block modification events to all of these triggers to make sure that they cannot modify claimed chunks without permission.

Probably not, unless someone depends on Wand Foci working through fake players in their claimed chunks.

### Other information
Add `runtimeOnlyNonPublishable('com.github.GTNewHorizons:ServerUtilities:2.4.10:dev')` to `dependencies.gradle` to test this change.
For testing, I recommend starting a dedicated server and two clients with different names, having one client claim some chunks (make sure that the team settings are set to "only allies" can break / place blocks) and have the other player try modifying those claimed chunks.

Also, I'm not sure if this is quite a bug or a feature, but I'm putting it in "bug fixes" anyway.

Theoretically, claimed chunks can also prevent people from attacking entities within that chunk. This PR doesn't try to add events for that, but I can if need be.

### Checklist


Nikolay-Sitnikov:
Wait just one moment, I forgot Wand Focus: Portable Hole!

src/main/java/dev/rndmorris/salisarcana/lib/PlayerHelper.java
```diff
@@ -40,4 +53,61 @@ public static ItemStack[] getItemsInInventory(EntityPlayer player, Class<? exten
...
+        final String name = profile.getName();
+        final UUID uuid = profile.getId();
+
+        if (name != null) {
+            ownerInfo.setString("username", name);
+        }
+        if (uuid != null) {
+            ownerInfo.setLong("uuidUpper", uuid.getMostSignificantBits());
+            ownerInfo.setLong("uuidLower", uuid.getLeastSignificantBits());
+        }
```
Nikolay-Sitnikov, line 92: If anyone has any objections to how the UUID is stored (such as if you want them to be stored in one `NBTTagList` instead), now is the time.

rndmorris: Perhaps `"UUIDMost"` and `"UUIDLeast"` to match how they're stored in `Entity.java`?

Nikolay-Sitnikov:
Turns out that Portable Holes are predictively executed on the client, which means that cancelling parts of them is going to be obnoxiously difficult. I am working on it.

